PR #8, opened 2016-12-29
Fix some data_type column migrations

spl:
I was unable to run the schema migrations from beginning to end due to the following problems:

1. `1_root` required superuser privilege to run.
2. `20150728-void_type` would not run in a transaction.

This PR fixes both. Since they are very similar (changing the values of the enum `data_type`), the migrations look similar.


dylex:
This looks good to me -- really just stuff I was too lazy to do at the time for early migrations.

makslevental:
@dylex should we merge this in?

dylex:
Completely up to you, @makslevental.  I would first use `./runsql check` to test it but if that passes I'd be fine with it.

makslevental:
thanks @spl for the fix.

PR #9, opened 2016-12-29
Change pgDBPort default configuration

spl:
Currently, the hardcoded socket is used if no host, port, or socket are specified in the config file. I think it is unlikely that this hardcoded socket would be used without specification. (I ran into this problem myself.) Therefore, I'm changing the way the host and port defaults are determined.


dylex:
I can see arguments either way for this one.  Really the config example should just be more explicit about the defaults, whatever they're chosen to be.  Regardless, if someone accepts this PR, you should the configuration on the production machines first: I know they're using socket connections and may be relying on the default, in which case this will break them.

spl:
@dylex True, defaults _are_ somewhat arbitrary. But I think it's far more likely that `localhost:5432` will be useful than `/tmp/.s.PGSQL.5432`. The former is rather standard while the latter is configuration-dependent. Production should definitely be using a config file rather than the defaults.

PR #38, opened 2017-07-17
add some comments'

makslevental:


kanishka-azimi:
I'll probably merge and delete the branch soon. These comments could only help.

makslevental:
@kanishka-azimi be aware they are not the most rigorous explanations - in fact some are probably wrong and therefore misleading.

PR #50, opened 2017-09-28
Remove selector abstraction - Part 2 - start removing Joins (rebased)

kanishka-azimi:
Passed most tests.

Uploading an mp4 seems to not reach end of the conversion process. Repeat test with tracing.


kanishka-azimi:
On dev2 next

PR #62, opened 2017-10-01
B2 - Add library, minimal test-suite - branch off dev

kanishka-azimi:
```
cabal configure --enable-tests
cabal install --only-dependencies --enable-tests
```


kanishka-azimi:
Later

PR #64, opened 2017-10-01
B1 - Use cabal version instead of git version - branch off dev

kanishka-azimi:
- remove Setup.Git module used to fetch and set package version
- remove confhook which used to set version and exe suffix if devel flag not set
- change dev script to extract version and cp exes after install (should delete, create symlink instead of copy file)
- fix cabal file to use version matching tag


kanishka-azimi:
Check version in bottom of page

kanishka-azimi:
A bunch of places use showVersion. Can simplify those later with simple module instead paths_databrary, generated by shell script "bumpVersion"

kanishka-azimi:
Add new command-line arg --version, which dumps showVersion for now

Add tag version:
- ensure branch is master
- ensure git status is clean
- git tag w/version
- git push tags

PR #65, opened 2017-10-01
B3 - Test some fixtures - branch off dev

kanishka-azimi:
Use json instead, prepare for post requests

### DIFF
--- a/databrary.cabal
+++ b/databrary.cabal
@@ -59,6 +59,91 @@ Flag sandbox
   default: False
   manual: True
 
+Library
+  exposed-modules: Data.RangeSet.Parse
+                 , Databrary.Store.AV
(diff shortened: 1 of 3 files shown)


kanishka-azimi:
Integration test bootstrapping:
- start writing fixtures, guess inputs needed (use ingest as a guide)
- start testing posts, check status
- interleave with gets, testing post + get
- also interleave with post "action" routes
- build up db, solr, store setup and teardown as add more tests
- expand fixtures, test variations for critical areas
- test register and login routes last, if ever
...
- refactor into fewer int tests and more unit tests, mock more dependencies (and test dependencies in isolation) start deleting tests in favor of type level checks

kanishka-azimi:
Later

PR #75, opened 2017-10-05
B1 - Use cabal version instead of git version

kanishka-azimi:
- set a real version in cabal file
- stop looking up git info, and stop overriding cabal version and exe suffix
- use shell script to copy exe into version suffixed exe


dev
```diff
@@ -19,20 +19,24 @@ cabal configure --user
 #cabal build
 
 echo "========Build Setup script into dist/setup/setup==============================================="
-echo "========Determine version from git commit======================================================"
```
kanishka-azimi: repeat this change above

PR #91, opened 2017-10-19
Automate staging workflow steps, similar to dev

kanishka-azimi:
FYI @muji786


kanishka-azimi:
Should check for presence of databrary.conf for build user as well

PR #95, opened 2017-10-19
Invoke databrary -w directly, copy assets from script

kanishka-azimi:
- Remove behavior from cabal, perform with wrapper shell script
- Also remove prior commented out code


kanishka-azimi:
FYI @muji786 read briefly, just so a second person has seen this and finds it understandable

PR #107, opened 2017-10-26
Use shell instead of Haskell custom build hook to copy, set transcode…

kanishka-azimi:
… scripts executable

- move towards simple build

Output in data directory:
```
-rwxr-xr-x. 1 kanishka kanishka  2550 Oct 26 17:48 transcode
-rwxr-xr-x. 1 kanishka kanishka  1886 Oct 26 17:48 transctl.sh
```

FYI: @muji786 , so one other person is aware of this change having occurred.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -35,11 +35,13 @@ run verb desc lbi cmd args = do
     : env
 
 -- chmod +x on transctl.sh and transcode scripts
+{-
 fixPerms :: PackageDescription -> LocalBuildInfo -> CopyDest -> IO ()
 fixPerms desc lbi copy = do
   setFileExecutable (dir </> "transctl.sh")
   setFileExecutable (dir </> "transcode")
   where dir = datadir $ absoluteInstallDirs desc lbi copy
+-}
 
 main :: IO ()
 {- defaultMainWithHooks does the default build things but also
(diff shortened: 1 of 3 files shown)


dev
```diff
@@ -29,7 +29,9 @@ built_exe=`ls -t $exe_dir/databrary-* | head -1`
 databrary_datadir="." $built_exe -w
 pushd web
 cp all.min.css all.min.css.gz all.min.js all.min.js.gz constants.json constants.json.gz $data_outputwebdir/
-popd web
```
kanishka-azimi: the popd command was wrong before

Setup.hs
```diff
@@ -80,11 +82,13 @@ main = defaultMainWithHooks simpleUserHooks
...
     postBuild simpleUserHooks args flag desc lbi
 -}
+{-
     -- do the chmod +x thing to transcode and transctl.sh
   , postCopy = \args flag desc lbi -> do
     fixPerms desc lbi (fromFlag $ copyDest flag)
     postCopy simpleUserHooks args flag desc lbi
   , postInst = \args flag desc lbi -> do
     fixPerms desc lbi NoCopyDest
     postInst simpleUserHooks args flag desc lbi
```
kanishka-azimi: I understand "copy" to mean the copying to the ~/.cabal/..../share/databrary-1 or similar. I suspect the copy hook is used to work for when development use to run with datadir=., but I don't know for sure.

PR #121, opened 2017-10-30
Create unit test suites that match the future cabal libs

kanishka-azimi:
- test-suites for env, mail, authent, solr, store, rename unit-test to dbrary
- for now include the under test source in the test suite, later use libs


kanishka-azimi:
Experience race conditions sometimes when building for the permission enum type fetch from db. can fix later.

PR #194, opened 2017-12-06
Front end build (generate.sh) run without requiring backend to start/build; use gzip command

kanishka-azimi:
Notes: e-tag computation disabled, extension is converted to mime-type on each fetch of static web asset instead of when the server starts up.

cc: @myicicle I might merge this next week. Please try this out some time if you get a chance.


kanishka-azimi:
Reopen later.

PR #196, opened 2017-12-08
Add ghci scripts to perform ghci steps in batch, main avoids exitSuccess

kanishka-azimi:
- exitSuccess was caused the ghciscript to stop early, so change to return ()

CC: @myicicle some convenience scripts. will update readme soon


kanishka-azimi:
Had to run main -w before attempting reload to avoid some strange ghci error in reload-gen-start

PR #237, opened 2017-12-22
Download original - volume level button, fix bugs

kanishka-azimi:
- expose button for download original
- fix long files names to only cutoff name, not extension
- for any file, use the current release value, not the release value at time of upload
- fix volume level download original to properly retrieve old file, instead of accidentally retrieving old containers with latest files


kanishka-azimi:
CC: @muji786 Just informing a second person this code changed.

PR #244, opened 2018-01-09
Aggressively guard any routes against public restricted access - repeat

kanishka-azimi:
To reach a first implementation for use on real volumes, guard most ways to retrieves parts of a volume with not found response.


kanishka-azimi:
I did a fairly deep search through the code for all areas that require read only access to a volume, asset, asset segment, slot/container, or record. All areas seem guarded appropriately.

kanishka-azimi:
Some repetition in the way the access is computed, refactor later to use a shared function.

PR #270, opened 2018-01-23
Conditionally include map data script tag in homepage

kanishka-azimi:
When user is entering the site through the homepage, include map data script tag.

When user enters the site through any other direct url besides the homepage, don't include map data script tag.


kanishka-azimi:
This almost works, but fails when you enter the site directly through another url, and then attempt to visit home page.

PR #271, opened 2018-01-24
Re-implement download session to restore bounded memory usage

kanishka-azimi:
Most likely upgrading from haskell compiler 7.10.3 to 8 introduced a change in behavior that broke an optimization that our homegrown zip generation logic relied upon

Use third party libraries to generate zip files in memory for download session, restoring the same or better memory behavior as old version of code. `zip-stream` appears to be one of the only libraries that offer this functionality. (There is also an in progress implementation of the feature for `zip` here - https://github.com/mrkkrp/zip/issues/20)

https://github.com/databrary/databrary/pull/269/files

### DIFF
--- a/Databrary/Action/Response.hs
+++ b/Databrary/Action/Response.hs
@@ -47,8 +47,11 @@ instance ResponseData BSL.ByteString where
 instance ResponseData BS.ByteString where
   response s h = responseBuilder s h . BSB.byteString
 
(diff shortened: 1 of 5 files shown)


kanishka-azimi:
If regression tests pass, remove dead code and merge.

kanishka-azimi:
Working partially, hit dead-end for now. Switching to alternative library.

PR #272, opened 2018-01-25
Attempt two at download session zip, using fork of zip library

kanishka-azimi:
- hoping building zip with ephmeral file handle will allow for computing size accurately to be able to populate content-length in http response and satisfy OS X archiver

attempt 1: https://github.com/databrary/databrary/pull/271
using https://github.com/mrkkrp/zip/pull/22/files

### DIFF
--- a/Databrary/Controller/Zip.hs
+++ b/Databrary/Controller/Zip.hs
@@ -114,15 +114,9 @@ containerZipEntry isOrig c l = do
 
 containerZipEntry2 :: Bool -> Container -> [AssetSlot] -> ActionM (ZIP.ZipArchive ())
 containerZipEntry2 isOrig c l = do
-  -- req <- peek
   let containerDir = makeFilename (containerDownloadName c) <> "/"
   zipActs <- mapM (assetZipEntry2 isOrig containerDir) l
   return (sequence_ zipActs)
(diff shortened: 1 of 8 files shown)


Databrary/Controller/Zip.hs
```diff
@@ -137,6 +194,28 @@ zipResponse n z = do
     , (hContentLength, BSC.pack $ show $ sizeZip z + fromIntegral (BS.length comment))
     ] (streamZip z comment)
 
+zipResponse2 :: BS.ByteString -> ZIP.ZipArchive () -> ActionM Response
+zipResponse2 n zipAddActions = do
+  req <- peek
+  u <- peek
+  let comment = BSL.toStrict $ BSB.toLazyByteString
+        $ BSB.string8 "Downloaded by " <> TE.encodeUtf8Builder (partyName $ partyRow u) <> BSB.string8 " <" <> actionURL (Just req) viewParty (HTML, TargetParty $ partyId $ partyRow u) [] <> BSB.char8 '>'
+  let temporaryZipName = "/tmp/placeholder.zip" -- TODO: generate temporary name for extra caution?
```
kanishka-azimi: Running out of disk space when using /tmp, need to use the "temp" location generally used by databrary.

PR #276, opened 2018-01-30
Fix Excerpt Thumbnail and Download Access for Partial Volume Level Sharing (attempt 1) - repeat

kanishka-azimi:
Make partially shared volumes behave like public volumes for assets where there is an excerpt present.

### DIFF
--- a/Databrary/Controller/Asset.hs
+++ b/Databrary/Controller/Asset.hs
@@ -75,20 +75,32 @@ import Databrary.View.Asset
 
 import Control.Monad.IO.Class
 
-getAsset :: Permission -> Id Asset -> ActionM AssetSlot
-getAsset p i = do
+getAsset :: Permission -> Bool -> Id Asset -> ActionM AssetSlot
+getAsset p checkDataPerm i = do
   mAssetSlot <- lookupAssetSlot i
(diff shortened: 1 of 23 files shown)


Databrary/Controller/Asset.hs
```diff
@@ -313,20 +323,22 @@ deleteAsset = action DELETE (pathAPI </> pathId) $ \(api, ai) -> withAuth $ do
...
 downloadOrigAsset = action GET (pathId </> pathSegment </< "downloadOrig") $ \(ai, seg) -> withAuth $ do
-  a <- getOrigAsset PermissionPUBLIC ai
+  a <- getAsset True PermissionPUBLIC True ai
   inline <- peeks $ lookupQueryParameters "inline"
   serveAssetSegment (null inline) $ newAssetSegment a seg Nothing
 
 thumbAsset :: ActionRoute (Id Asset, Segment)
 thumbAsset = action GET (pathId </> pathSegment </< "thumb") $ \(ai, seg) -> withAuth $ do
-  a <- getAsset PermissionPUBLIC ai
+  a <- getAsset False PermissionPUBLIC True ai
```
kanishka-azimi: Should be check false

PR #277, opened 2018-01-31
Fix Excerpt Thumbnail and Download Access for Partially Shared Volumes - Backend Phase 1

kanishka-azimi:
At this point, the backend will allow access to an excerpt's video segment only by for partially shared volumes when public users are viewing.

Next steps:
- fix back end to always provide the correct segment when a segment is requested
- fix front end to only attempt to access the excerpt segment of a video in volume overview (and search?)

cc: @muji786

### DIFF
--- a/Databrary/Model/AssetSegment/Types.hs
+++ b/Databrary/Model/AssetSegment/Types.hs
@@ -95,7 +95,7 @@ instance Has (Maybe Release) AssetSegment where
   view AssetSegment{ segmentAsset = a, assetExcerpt = Just e } = excerptRelease e <> view a
   view AssetSegment{ segmentAsset = a } = view a
 instance Has Release AssetSegment where
-  view = view . (view :: AssetSegment -> Maybe Release)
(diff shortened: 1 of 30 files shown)


Databrary/Model/AssetSegment/Types.hs
```diff
@@ -91,12 +97,26 @@ instance Has Format AssetSegment where
...
+           fold (
+                excerptRelease e  -- Maybe Release monoid takes the first just, if both just, then max of values
+             <> getAssetSlotReleaseMaybe a) -- TODO: should I expose the guts of getAssetSlotRelease2?
+      in 
+        EffectiveRelease {
+          effRelPublic = rel
+        , effRelPrivate = rel
+        }
+    AssetSegment{ segmentAsset = a } ->
+      getAssetSlotRelease2 a
```
kanishka-azimi: ensure effectiverelease is computed correctly for all possibly combinations

PR #281, opened 2018-02-02
Kanishka dbd627 rebase

kanishka-azimi:


kanishka-azimi:
Can read data should embed a call to checkpermission, or reuse it's implementation, to ensure both are run together

PR #288, opened 2018-02-05
Replace TH with hardcoded format list

chreekat:
List taken from querying local db.

List formatted with `brittany` - not saying it's the prettiest. :)


kanishka-azimi:
cc: @muji786 (:

PR #290, opened 2018-02-06
Start adding tracing to serveAssetSegment - develop

kanishka-azimi:


Databrary/Store/AssetSegment.hs
```diff
@@ -58,6 +58,10 @@ genVideoClip _ src (Just clip) _ dst | Nothing <- Range.getPoint clip = do
   dstfp <- case dst of
     Left _ -> return "-"
     Right rp -> unRawFilePath rp
+  print ("about to slice video file")
+  let upperBoundArgs = maybe [] (\u -> ["-t", sb $ u - lb]) ub
```
kanishka-azimi: @theowolf you might find this part of the code interesting, this is the underlying ffmpeg usage that extracts a segment of a video file

PR #294, opened 2018-02-06
Version CSS and JS files

kanishka-azimi:
Compute a hash, use it when fetching js and css files.

cc: @chreekat

### DIFF
--- a/Databrary/View/Angular.hs
+++ b/Databrary/View/Angular.hs
@@ -25,11 +25,14 @@ import Databrary.View.Template
 ngAttribute :: String -> H.AttributeValue -> H.Attribute
 ngAttribute = H.customAttribute . H.stringTag . ("ng-" <>)
 
-webURL :: BS.ByteString -> H.AttributeValue
+webURL :: BS.ByteString -> H.AttributeValue -- TODO: stop using this?
 webURL p = actionValue webFile (Just $ StaticPath p) ([] :: Query)
 
-htmlAngular :: [WebFilePath] -> [WebFilePath] -> BSB.Builder -> RequestContext -> H.Html
-htmlAngular cssDeps jsDeps nojs auth = H.docTypeHtml H.! ngAttribute "app" "databraryModule" $ do
+versionedWebURL :: BS.ByteString -> BS.ByteString -> H.AttributeValue
+versionedWebURL version p = actionValue webFile (Just $ StaticPath p) ([(version,Nothing)] :: Query)
+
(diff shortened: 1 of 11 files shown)


Databrary/View/Angular.hs
```diff
@@ -25,11 +25,14 @@ import Databrary.View.Template
 ngAttribute :: String -> H.AttributeValue -> H.Attribute
 ngAttribute = H.customAttribute . H.stringTag . ("ng-" <>)
 
-webURL :: BS.ByteString -> H.AttributeValue
+webURL :: BS.ByteString -> H.AttributeValue -- TODO: stop using this?
```
kanishka-azimi: stop using it, or combine the two functions to take a `Maybe BS.ByteString`

Databrary/Web/Types.hs
```diff
@@ -33,6 +33,7 @@ data Web = Web
     MVar
 #endif
       WebFileMap
+  , webVersion :: BS.ByteString -- Compute hash or increment of all js and css files
```
kanishka-azimi: this may be repeating information that is computed in he WebFileMap, but I am hoping we remove the WebFileMap in a while and only this remains.

kanishka-azimi:
@chreekat  I had to merge this before your work since this was a little more urgent than usual. It shouldn't be too hard to rebase into your work.

databrary.nix
```diff
@@ -84,6 +85,7 @@ mkDerivation rec {
 
     cp messages.conf volume.json $data_outputdir
 
+    cp jsCssVersion.txt $data_outputdir
     mkdir $data_outputdir/web
     cd web
     cp all.min.css all.min.css.gz all.min.js all.min.js.gz constants.json constants.json.gz $data_outputdir/web
```
kanishka-azimi: Not using constants.json as an input. I am not sure that constants.json is referenced anywhere directly, or if it is embedded in all.min.js.

Databrary/Web/Rules.hs
```diff
@@ -98,4 +99,7 @@ generateAll = do
 
 generateWebFiles :: IO WebFileMap
 generateWebFiles = do
-  execStateT (either fail return =<< runExceptT generateAll) HM.empty
+  fm <- execStateT (either fail return =<< runExceptT generateAll) HM.empty
+  -- TODO: variables for filenames
+  callCommand "cat web/all.min.js web/all.min.css | md5sum | cut -d ' ' -f 1 > jsCssVersion.txt"
```
kanishka-azimi: @myicicle For future reference, this is the unix command used to compute the new hash. We may want to add more files as input, if we discover that is needed. This should be sufficient though.

PR #295, opened 2018-02-06
WIP: Use src/, app-*/, and cbits/

chreekat:
Required a few small changes along the way.

Also deleted the (unused?) Databrary.Controller.Email, which doesn't
compile.

Also, as a first pass I just copied all deps from the old `executable: databrary` into the new library section. Presumably the new executable section will be able to ditch a lot of them, though it doesn't truly matter.


chreekat:
I wouldn't merge this yet - I think it definitely caused a big slowdown in ghci-databrary start time. Maybe loading the exe component caused the lib component to switch from byte code to object code (a lot slower). I'll look into it.

kanishka-azimi:
Yeah, I usually use a library and executable, using cabal build and cabal test mostly. Since working on this project inside of nix, I have used ghci and lived with the executable only to start.

The key constraint is that ghci has to stay relatively fast, because front end builds have to use it somewhat often. I wouldn't mind if it slowed down for just back end development.

kanishka-azimi:
app-generate/generate.hs - check hot reload

kanishka-azimi:
```
++++ databrary_datadir=.
++++ dist/build/databrary/databrary -w
LOG:  database system is shut down
dist/build/databrary/databrary: error while loading shared libraries: libHSdatabrary-1-B3SeHsVxdy2KXEA9J4wl8M-ghc8.0.2.so: cannot open shared object file: No such file or directory
+ exitHandler
```
from `nix-build --attr databrary` on build server (fresh clone of develop)

kanishka-azimi:
I would back out the library section change and use your new directory layout but with the old single executable, see if we can get that mergeable for now.

kanishka-azimi:
Once we have everything compiling with nix-build, we should do a full regression test of the back end routes as well. I can show you what I do currently... I was going to shield you from having to be aware of all the places to regression test for a while, but its not a big deal to bump that up sooner.

kanishka-azimi:
Can the cbits use subdirectories? If so, it would be nice to repeat the old structure of databrary/store

kanishka-azimi:
Note to self: patching appeared to still be running on some files, unsure.

PR #304, opened 2018-02-07
Cabal cleanup

kanishka-azimi:


kanishka-azimi:
Note to self: ~ 14 mins build time before optimizing again

PR #312, opened 2018-02-13
nix build including tests

chreekat:
Includes a number of improvements - see commits


chreekat:
Note this PR is to develop. nix-build works but we should make sure the assets load ok. :)

databrary.cabal
```diff
@@ -470,7 +469,7 @@ Executable databrary
   default-language: Haskell2010
   default-extensions: MultiParamTypeClasses, FlexibleContexts, FlexibleInstances, ScopedTypeVariables, ConstraintKinds, PatternGuards
 
-  ghc-options: -threaded -Wall -fwarn-tabs -funbox-strict-fields
```
chreekat: -fwarn-tabs is on by default; -funbox-strict-fields is ["a big hammer" and mostly unnecessary](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-optimisation.html?highlight=unbox-st#ghc-flag--funbox-strict-fields).

PR #318, opened 2018-02-15
Chreekat dbd 752  test repl

kanishka-azimi:


databrary.cabal
```diff
@@ -530,11 +533,365 @@ test-suite databrary-test
...
+      xml >= 1.3.14,
+      zlib >= 0.6.1.1,
+      range-set-list >= 0.1.2.0,
+      invertible >= 0.1.1,
+      web-inv-route >= 0.1,
+      smtp-mail >= 0.1.4.6,
+      zip >= 0.1.3,
+      conduit-combinators >= 1.1.1,
+      binary >= 0.8.3.0,
+      path >= 0.5.13,
```
kanishka-azimi: Databrary.nix testdepends can be ignorant of this probably?

kanishka-azimi:
I thought further about my proposal of bringing all tests into library. I like having a little delay before when I have to get the tests compiling after modifying code, so I will probably withdraw that proposal. Although it might teach us some extra discipline to make the tests light and not too cumbersome as well as getting us fully into test driven rather than test trailing. Just not sure if test driven should be a goal or if test trailing is a better approach.

reload-gen-start
```diff
@@ -1,5 +1,4 @@
-:module + Databrary.Main
-:main -w
+:module + Main
+:main
 :reload
-:main -w
-main
+:main
```
kanishka-azimi: this and above were temporary hacks... reminder to ensure the old and scripts remain and you duplicate them for testing `ghci-test-databrary`

test/Test.hs
```diff
@@ -1,9 +1,27 @@
...
 tests :: TestTree
-tests = testGroup "databrary" []
+tests = testGroup "databrary" [testMain]
+
+testMain :: TestTree
+testMain = testGroup
+    "Databrary.Main"
+    [ expectFail $ testCase
+          "flagConfig sanity check"
+
```
kanishka-azimi: this is a good test of loading the top most module, but if you are in a similar situation in the future and you want to see if databrary compiles, I would focus first on the lowest levels such as avProbe and usage of cracklib, to isolate any errors linking to c libraries early, then work up to a test like this which triggers loading all linked libraries.

kanishka-azimi:
If you like this approach, let's merge this before master drifts.

PR #327, opened 2018-02-18
Remove compile time fetching of db enum values

kanishka-azimi:
One step closer to no IO / DB connections at compile time.

We can test directly on staging. With all the compile time checks, the chance of anything breaking is close to nothing and rolling back is very easy. 

cc: @chreekat


kanishka-azimi:
I would shift towards using a lookup table instead of enum, and loading the lookups on startup.

chreekat:
> I would shift towards using a lookup table instead of enum, and loading the lookups on startup.

But don't early-optimize...

PR #328, opened 2018-02-19
Expand insert / update queries in Funding to underlying queries

kanishka-azimi:
- continuing to reach for no db queries at compile time.

cc: @chreekat


kanishka-azimi:
Exercised add, insert, update, delete funding on staging.

PR #330, opened 2018-02-19
Expand Release to underlying queries

kanishka-azimi:
- used dumped splices output to remove compile time db connections in Release module


kanishka-azimi:
Exercised insert, update, delete release for a session container.

PR #332, opened 2018-02-21
Expand Measure related SQL logic to underlying queries

kanishka-azimi:
cc: @chreekat  - this is probably the fastest hack to expand any SimpleQuery usages


kanishka-azimi:
Will test in the morning.

kanishka-azimi:
Tested add measure, update measure, remove measure

PR #333, opened 2018-02-21
Add back model.release.sql because containers uses it

kanishka-azimi:
cc: @chreekat


src/Databrary/Model/Release/SQL.hs
```diff
@@ -0,0 +1,47 @@
...
+
+import qualified Language.Haskell.TH as TH
+
+import Databrary.Model.SQL.Select
+import Databrary.Model.Audit.SQL
+import Databrary.Model.Slot.SQL
+
+releaseRow :: Selector -- ^ @'Release'@
+releaseRow = selectColumn "slot_release" "release"
+
```
kanishka-azimi: delete this when container model is removed

PR #343, opened 2018-02-24
Add Lab Nanny and Integration Test packages

kanishka-azimi:
Create new top level directory, might eventually house all source packages.
Create corresponding runtime directory for integration test package.


kanishka-azimi:
Feel free to propose improvements. @chreekat @theowolf

PR #345, opened 2018-02-24
Expand Asset Revision simple queries

kanishka-azimi:
Expand replace asset and asset is replaced queries to dumped splices

Test Monday

cc: @chreekat


kanishka-azimi:
tested

PR #348, opened 2018-02-26
Cherry picked documentation and tests from work in progress metadata upload

kanishka-azimi:
- add type signatures and more verbose variable names
- add unit test and notes for future unit tests

Trimming down work in progress branch.

cc: @chreekat

### DIFF
--- a/src/Databrary/Action/Form.hs
+++ b/src/Databrary/Action/Form.hs
@@ -7,18 +7,25 @@ import qualified Data.Map.Strict as Map
 import Data.Maybe (fromMaybe)
 import Data.Word (Word64)
 import qualified Network.Wai as Wai
+import Network.Wai.Parse (FileInfo)
 
 import Databrary.Has (peeks)
 import Databrary.HTTP.Form.Data
 import Databrary.HTTP.Parse
 import Databrary.Action.Types
+import qualified Databrary.JSON as JSON
 
 getFormData :: FileContent a => [(BS.ByteString, Word64)] -> ActionM (FormData a)
-getFormData fs = do
-  f <- peeks $ FormData . Map.fromList . Wai.queryString
-  c <- parseRequestContent (fromMaybe 0 . (`lookup` fs))
+getFormData fileLimits = do
+  (mkFormData
+   :: Map.Map BS.ByteString BS.ByteString -> Maybe JSON.Value -> Map.Map BS.ByteString (FileInfo a) -> FormData a)
+    <- peeks $ (\httpReq -> (FormData . Map.fromList . Wai.queryString) httpReq)
+  (c :: Content a) <- parseRequestContent getFileMaxSizeByFieldName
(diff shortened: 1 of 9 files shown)


src/Databrary/Action/Form.hs
```diff
@@ -7,18 +7,25 @@ import qualified Data.Map.Strict as Map
...
 import Databrary.HTTP.Parse
 import Databrary.Action.Types
+import qualified Databrary.JSON as JSON
 
 getFormData :: FileContent a => [(BS.ByteString, Word64)] -> ActionM (FormData a)
-getFormData fs = do
-  f <- peeks $ FormData . Map.fromList . Wai.queryString
-  c <- parseRequestContent (fromMaybe 0 . (`lookup` fs))
+getFormData fileLimits = do
+  (mkFormData
```
chreekat: +1 mkFormData, -1 to its huge (and scoped) type sig. Ok for temp cleanup I guess.

chreekat: I'm not a fan of eta abstraction (explicit function arguments) in the cases where they are easy to avoid. In this particular case it's pure noise

src/Databrary/Controller/Party.hs
```diff
@@ -125,12 +125,17 @@ processParty api p = do
     orcid <- "orcid" .:> deformNonEmpty (deformRead blankORCID)
     affiliation <- "affiliation" .:> deformNonEmpty deform
     url <- "url" .:> deformNonEmpty deform
-    avatar <- "avatar" .:>
-      (maybe (deformOptional $ return Nothing) (\a -> do
-        f <- deformCheck "Must be an image." formatIsImage =<<
-          deformMaybe' "Unknown or unsupported file format."
-          (getFormatByFilename (fileName a))
-        return $ Just $ Just (a, f)) =<< deform)
+    (avatar :: (Maybe (Maybe (FileInfo TempFile, Format)))) <- "avatar" .:> do
```
chreekat: Ok, if Deform needs lots of `(Maybe (Maybe ...))`, I agree it needs work.

test/Databrary/HTTP/Form/DeformTest.hs
```diff
@@ -0,0 +1,71 @@
...
+
+import Databrary.HTTP.Form.Deform
+
+tests :: TestTree
+tests = testGroup "Databrary.HTTP.Form.DeformTest"
+    [ testCase "textInteger-1" (textInteger "1" @?= Right (1::Int))
+    
+    ]
+
+-- usages:
```
chreekat: I'm not a fan of code clutter. I'd rather put this whole thing in an Issue. This may be a difference of opinion, not saying one way is necessarily better.

kanishka-azimi:
Will keep those preferences in mind. I can agree with some, others we can duke out in future PRs!

PR #353, opened 2018-02-27
Kanishka expand queries 226

kanishka-azimi:
Test these + Lookupvolumeasset Thursday

cc: @chreekat


kanishka-azimi:
Skipped testing lookupVolumeAsset. Should be fine. Will discover any discrepancies when testing ingest.

Others will be tested with regression test suite.

PR #354, opened 2018-02-27
Add additional metrics, keep dynamic mapping

kanishka-azimi:


src/Databrary/Controller/Ingest.hs
```diff
@@ -46,13 +71,114 @@ postIngest = multipartAction $ action POST (pathId </< "ingest") $ \vi -> withAu
...
+         Just idCol ->
+             case HMP.lookup (TE.encodeUtf8 idCol) csvRecord of
+                 Just fieldVal ->
+                     -- getCSVField row field
+                     print ("save measure", "ID", fieldVal)
+                 Nothing ->
+                     pure () -- TODO: error, impossible?
+         Nothing ->
+             pure () -- field isn't used by this volume, so don't need to save the measure
+     pure ()
```
kanishka-azimi: Two version of this function. 1. extract and return sample; 2. extract and run as above.

src/Databrary/Controller/Ingest.hs
```diff
@@ -46,13 +71,114 @@ postIngest = multipartAction $ action POST (pathId </< "ingest") $ \vi -> withAu
...
+runImport records mapping =
+    mapM
+        (\record -> createRecord mapping record)
+        records
+
+createRecord :: ParticipantFieldMapping -> CSV.NamedRecord -> IO () -- TODO: error or record
+createRecord mapping csvRecord = do
+     -- record <- makeRecord
+     -- for each field
+     case pfmId mapping of --     metricId = getId name
```
kanishka-azimi: move this logic out to shared function

kanishka-azimi:
Delete this branch as soon as no one else is using it.

PR #369, opened 2018-03-08
Fix improper seq val for party in development database creation

kanishka-azimi:
- a few months ago, "Testarosa" was added, but we left the sequence at 6
- incrementing to proper value should avoid error on first registration on an empty db
cc: @chreekat


kanishka-azimi:
Tested, appears to work.

PR #383, opened 2018-03-13
Replace pgranges

chreekat:
Implements the one function used (segments_union(segment[], segment)) with an implementation in pgplsql. Removes obsolete deps from build.

* [x] make
* [x] make db
* [x] make devel
* [x] make nix-build


chreekat:
One sec, need to fix something in the Makefile

chreekat:
Fixed

pgranges/pgranges.sql
```diff
@@ -1,3 +0,0 @@
-CREATE OR REPLACE FUNCTION segments_union(segment[], segment[]) RETURNS segment[] IMMUTABLE STRICT LANGUAGE C AS 'pgranges.so', 'ranges_union';
-CREATE OR REPLACE FUNCTION segments_union(segment[], segment) RETURNS segment[] IMMUTABLE STRICT LANGUAGE C AS 'pgranges.so', 'ranges_union1';
-CREATE AGGREGATE "segments_union" (segment) (SFUNC = segments_union, STYPE = segment[], INITCOND = '{}');
```
kanishka-azimi: please document what happened to "aggregate"

databrary.nix
```diff
@@ -106,7 +108,7 @@ in
   '';
   postBuild = ''
     kill -INT `head -1 $socket_path/postmaster.pid`
-    ln -sf ${nodePackages.shell.nodeDependencies}/lib/node_modules node_modules
+    ln -sf ${nodePackages.shell.nodeDependencies}/lib/node_modules .
```
kanishka-azimi: Please document

databrary.nix
```diff
@@ -80,7 +82,7 @@ in
     base tasty tasty-expected-failure tasty-hunit
   ];
   executableToolDepends = [
-    postgresql.postgres
+    postgresql
```
kanishka-azimi: Note to self: this changed because we no longer have to expose multiple build outputs nested under a container.

pgranges/pgranges.sql
chreekat: I thought it was a dup from [0.sql](https://github.com/databrary/databrary/blob/chreekat/replace-pgranges/schema/0.sql#L322), but ... oh, yes, it is. The aggregate is loosely coupled to the function it is defined to use.

schema/20180313-segments_union.sql
```diff
@@ -0,0 +1,63 @@
...
+$$;
+
+-- and now, a test
+-- FIXME add this to the test suite
+/*
+create temporary table segs ( s segment );
+insert into segs values ('empty'), ('(4,5)'),('(0,1)'), ('(1,2)'), ('(2,3)'), ('(-1,0)'), ('[2,2]'), ('[6,7]'),('(5,7]'),('[6,8]'), ('empty');
+select ss.s = '{"(-1,0)","(0,1)","(1,3)","(4,5)","(5,8]"}' as "test passed"
+from (select segments_union(s) s from segs) as ss;
+*/
```
kanishka-azimi: Can I persuade you to update this to use VALUES instead of temporary table, per note from before?

chreekat: With the understanding that temp tables go away at the close of the session and shadow any existing tables on creation?

kanishka-azimi: yeah, just trying to minimize the hesitation to copy/paste to create more tests. very minor, either way

schema/20180313-segments_union.sql
```diff
@@ -0,0 +1,60 @@
+-- Implement segments_union(segment[], segment) in plpgsql. Remove the unused
+-- segments_union(segment[], segment[]).
+
+-- Fare thee well
+drop function if exists segments_union(segment[], segment[]);
+
+-- | Add a new segment to a list, merging overlapping or adjacent segments into
+-- one.
+--
+-- We assume acc (the input list) is sorted and non-overlapping.
```
kanishka-azimi: Future improvement: crash if not sorted

schema/20180313-segments_union.sql
```diff
@@ -0,0 +1,60 @@
...
+end
+$$;
+
+-- and now, a test
+-- FIXME add this to the test suite
+/*
+create temporary table segs ( s segment );
+insert into segs values ('empty'), ('(4,5)'),('(0,1)'), ('(1,2)'), ('(2,3)'), ('(-1,0)'), ('[2,2]'), ('[6,7]'),('(5,7]'),('[6,8]'), ('empty');
+select ss.s = '{"(-1,0)","(0,1)","(1,3)","(4,5)","(5,8]"}' as "test passed"
+from (select segments_union(s) s from segs) as ss;
```
kanishka-azimi: Can you use segments_union the function, not the aggregate once as well. Really confusing that they are named the same.

schema/20180313-segments_union.sql
chreekat: nm I figured out how to do it, needed a bit of syntax I wasn't aware of. Will do for the experience

schema/20180313-segments_union.sql
```diff
@@ -0,0 +1,60 @@
...
+        raise exception 'Expected one-dimensional state';
+    end if;
+    if isempty(seg) then
+        result := acc;
+    else -- seg is non-empty range
+        if acc_length = 0 then
+            result := ARRAY[seg];
+        else -- state is non-empty array
+            -- 1. Skip over strictly less-than, unadjacent segments
+            i := 1;
```
kanishka-azimi: can't remember if plpgsql allows nested blocks, can probably declare i , tmp here

schema/20180313-segments_union.sql
```diff
@@ -0,0 +1,60 @@
...
+            while acc[i] << seg and acc[i] -|- seg is false and i <= acc_length
```
kanishka-azimi: Maybe document the operator soup here.

schema/20180313-segments_union.sql
chreekat: I expanded the comment here; it might ameliorate some of these concerns (thought they're still somewhat valid).

schema/20180313-segments_union.sql
chreekat: Done

kanishka-azimi:
Maybe syntax is too new?
screenshot from 2018-03-13 15-09-58

schema/20180313-segments_union.sql
chreekat: I waffled on some of the the other tests that I added because they were easy to add, like ensuring that acc isn't multidimensional. Technically that can also just be assumed, for the same reason that we can assume the array is sorted.  :man_shrugging:

kanishka-azimi:
`ALTER USER databrary WITH SUPERUSER;` remove this from 00.sql

kanishka-azimi:
Drop database from all development servers and laptops; force creation of new

kanishka-azimi:
Rare error that will occur whenever we try to change postgres version in place on dev instances:
```
connection to postgres socket...
FATAL:  database files are incompatible with server
DETAIL:  The data directory was initialized by PostgreSQL version 9.6, which is not compatible with this version 9.5.9.
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/home/kanishka/dev/databrary/databrary-nix-db/work/.s.PGSQL.5432"?
```
Maybe we should make the postgres version part of the path to the database directory to prevent this in the future.

kanishka-azimi:
> Drop database from all development servers and laptops; force creation of new
Sometimes when we had actions like this in past work, we would add a temporary step to the "ghci-databrary" or "init-psql" script to force deletion of existing db and create new if some condition was true. Then remove that "migration" action after a few weeks.

PR #388, opened 2018-03-14
Two simple helper functions abstracting cassava details + tests

kanishka-azimi:
- getHeaders converts header names from cassava specific type to a more general type
- extractColumnDefaulting gets one column from a list of rows, adding default
to strip Maybe, since Nothing should never occur with known header names

one of many small PRs to come to merge full participant upload feature.


kanishka-azimi:
Merging myself for expediency.

PR #390, opened 2018-03-14
Add postgresql-simple pool

chreekat:
Rather than a pool of tupled connections, I want a tuple of pools. This does the thing.

Most of the noise is due to reverting the creation of tupled connections. The meat of the change is in initDB. Since all of the website only uses MonadDB / withDB, the change is very quiet otherwise.

### DIFF
--- a/databrary.cabal
+++ b/databrary.cabal
@@ -29,7 +29,7 @@ Flag sandbox
 library
   hs-source-dirs: src
   c-sources: src/Databrary/Store/avFrame.c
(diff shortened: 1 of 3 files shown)


chreekat:
Unless you had a strong reason to make a pool of tuples?

kanishka-azimi:
I had it mixed up.

chreekat:
Ok no prob. Let me rebase the conflict.

kanishka-azimi:
After rebase, start the app once, view search for "databrary", view the volume, traverse to a session, play a video. Then, create a new volume, add a session, upload a video. Then merge.

chreekat:
Done and done. ;)

PR #392, opened 2018-03-14
Move in helper functions for summarizing values in each column of a csv

kanishka-azimi:
cc: @chreekat Feel free to review and I will make adjustments based on your comments in upcoming PRs.

### DIFF
--- a/src/Data/Csv/Contrib.hs
+++ b/src/Data/Csv/Contrib.hs
@@ -1,19 +1,35 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Data.Csv.Contrib
   ( getHeaders
+  , extractColumnsDistinctSample
   , extractColumnDefaulting
+  , extractColumn
   ) where
 
 import qualified Data.ByteString as BS
 import qualified Data.Csv as CSV
 import qualified Data.HashMap.Strict as HMP
+import qualified Data.List as L
 import qualified Data.Vector as V
 import Data.Vector (Vector)
 
 getHeaders :: CSV.Header -> [BS.ByteString]
 getHeaders = V.toList
 
--- TODO: use in extractSampleColumns
+extractColumnsDistinctSample :: Int -> CSV.Header -> Vector CSV.NamedRecord -> [(BS.ByteString, [BS.ByteString])]
+extractColumnsDistinctSample maxSamples hdrs records =
+    zip hdrs'
+        ( fmap
+              ( getSample
+              . (\hdr -> extractColumnDefaulting hdr records))
+              hdrs' )
+  where
+    getSample :: [BS.ByteString] -> [BS.ByteString]
+    getSample vals =
+        (take maxSamples . L.nub) vals
+    hdrs' :: [BS.ByteString]
+    hdrs' = getHeaders hdrs
+
 extractColumnDefaulting :: BS.ByteString -> Vector CSV.NamedRecord -> [BS.ByteString]
 extractColumnDefaulting hdr records =
    extractColumn hdr records (maybe "" id)
--- a/src/Databrary/Model/Record.hs
+++ b/src/Databrary/Model/Record.hs
@@ -8,6 +8,7 @@ module Databrary.Model.Record
   , changeRecord
   , removeRecord
   , recordJSON
+  , columnSampleJson
 --  , recordJSONRestricted
   -- for testing only
   , extractParticipantFieldRows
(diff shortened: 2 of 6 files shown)


test/Data/Csv/ContribTest.hs
```diff
@@ -12,4 +12,14 @@ tests :: TestTree
...
+    , testCase "extractColumnsDistinctSample"
+        (extractColumnsDistinctSample
+           1
+           (V.fromList ["c1"])
+           (V.fromList
+              [ (HMP.fromList [("c1", "val1")])
+              , (HMP.fromList [("c1", "val1")])
+              , (HMP.fromList [("c1", "val2")])
+              ])
+           @?= [("c1", ["val1"])])
```
kanishka-azimi: @muji786 test driven is coming back strong...

src/Data/Csv/Contrib.hs
```diff
@@ -1,19 +1,35 @@
...
 import qualified Data.HashMap.Strict as HMP
+import qualified Data.List as L
 import qualified Data.Vector as V
 import Data.Vector (Vector)
 
 getHeaders :: CSV.Header -> [BS.ByteString]
 getHeaders = V.toList
 
--- TODO: use in extractSampleColumns
+extractColumnsDistinctSample :: Int -> CSV.Header -> Vector CSV.NamedRecord -> [(BS.ByteString, [BS.ByteString])]
```
kanishka-azimi: Should have used Map ByteString [ByteString], easier to combine or work with further later.

src/Databrary/Model/Record.hs
```diff
@@ -63,3 +71,20 @@ recordJSON publicRestricted r@Record{ recordRow = RecordRow{..}, ..} = JSON.Reco
...
+extractParticipantFieldRows :: [BS.ByteString] -> Vector CSV.NamedRecord -> [(BS.ByteString, [BS.ByteString])]
+extractParticipantFieldRows participantFieldHeaders records =
+    (zip participantFieldHeaders . fmap (\hdr -> extractColumnDefaulting hdr records)) participantFieldHeaders
+
+columnSampleJson :: BS.ByteString -> [BS.ByteString] -> JSON.Value
+columnSampleJson hdr sampleValues =
+    JSON.object [
+          "column_name" JSON..= hdr
+        , "samples" JSON..= sampleValues
+        ]
```
kanishka-azimi: should use type with tojson, will play nicer with servant eventually

src/Data/Csv/Contrib.hs
chreekat: Just use Foldable/Traversable at use sites, change types at your leisure.

src/Data/Csv/Contrib.hs
```diff
@@ -1,19 +1,35 @@
...
 
 import qualified Data.ByteString as BS
 import qualified Data.Csv as CSV
 import qualified Data.HashMap.Strict as HMP
+import qualified Data.List as L
 import qualified Data.Vector as V
 import Data.Vector (Vector)
 
 getHeaders :: CSV.Header -> [BS.ByteString]
 getHeaders = V.toList
```
chreekat: I don't like type synonyms for exactly the reason you (rightly) had to create this function.

src/Data/Csv/Contrib.hs
chreekat: This name and type sig combine into perfect incomprehensibility. I think more, specific types is the right kind of verbosity, allowing function names to be shorter without losing clarity. For instance, I have no idea if "sample" refers to the inputs or the outputs. But if there was a Sample somewhere between the arrows, it would be perfectly clear.

src/Databrary/Model/Record.hs
```diff
@@ -8,13 +8,21 @@ module Databrary.Model.Record
...
   , recordJSON
+  , columnSampleJson
 --  , recordJSONRestricted
+  -- for testing only
+  , extractParticipantFieldRows
   ) where
 
 import Control.Monad (guard)
+import qualified Data.ByteString as BS
+import qualified Data.Csv as CSV
```
chreekat: Bold re capitalization ;) 

I lean marginally towards sticking with whatever we're given. Camel casing acronyms seems to be the norm for Haskell, anyway.

src/Data/Csv/Contrib.hs
kanishka-azimi: It would be instructive to understand how you would have named this and what types you would have used. This is about a 4 out of 10 for what I would I aspire to. It was a function I had to add last minute, after the original design.

I tend to use a combination of typenames and variables names to document. I tried using type aliases in the past, but I didn't like the noise and indirection they introduced.

```
extractColumnsDistinctSample :: Int -> CSV.Header -> Vector CSV.NamedRecord -> [(BS.ByteString, [BS.ByteString])]
extractColumnsDistinctSample maxSamples hdrs records
```
would probably be clearest to me if it was
```
distinctSampleValuesFromEachColumn :: Int -> CSV.CSVData -> Map Text [BS.ByteString]
distinctSampleValuesFromEachColumn maxSamples csvData =
```
The type in cassava doesn't really provide a clean abstraction over headers + rows, which also introduces the messy fact that you have to lookup into a Map, even though you know the key is present.

kanishka-azimi:
> As a fierce opponent to dead code, I prefer to see utility functions like these built after the fact. I like to start at the top, building the business logic while stubbing out the little pieces with [placeholders]

I definitely use placeholder style when I am implementing. Currently, I am still in scrappy reorganize mode, then I will start getting fierce about dead code and compiler warnings again, and also focus on code and rush less.

I could use some help/advice on how to achieve a balance between minimum work in progress code (code on branches instead of in master) while also not adding speculative code. The PRs yesterday and today are mostly cutting up work from prior weeks, not that is stabilizing....

My favorite workflow is:
- break of small pieces of a feature
- start implementing, then opportunistically pick pieces to merge that are considered complete (and attempt to expose to user if possible). Aim for as many small merges as you can safely reach, while also having agreement from product owner that the piece being merge aligns closely with requirements.

To decrease the amount of time that our frontend development has to wait until it can start building demos and getting our product owner actively using/testing new code, I have relented a bit on my focus on constant merges. My current workflow is:
- break of sizeable chunk of a feature
- race to implement minimal pieces of backend to allow some front end development and real testing to proceed, adding lots of TODOs/FIXMEs in code
- once acceptance testing/requirements for some chunk nears maturity, revise large branch and carve up into small branches that one can merge

I am in the last phase currently and I am trying to merge as much as possible that seems mature. The most mature parts are the inner modules...

src/Data/Csv/Contrib.hs
chreekat: Names don't exist in a vacuum; knowing how a named thing is used is just as important as what the name refers to. I think we all dislike super long names, but I think we can be a little more specific, and say there is a relation

```
             Frequency of use
Length ⁓ ————————————————————————
           Generality of function
```

If a function is only used once, perhaps the best name for it is `f` (or no name at all). If it's very general, it should likewise be short (e.g. `map`).

> I tried using type aliases in the past, but I didn't like the noise and indirection they introduced.

Synonyms bad, newtypes good. That's how I feel about it. I usually end up unwrapping it right away, like

```haskell
newtype Whatev = Whatev Int
foobar :: Whatev -> Bool
foobar (Whatev i) = ...
```

so implementation isn't really any noisier. The upside is in the Haddocks and in giving one extra thought when using the function.

chreekat: > The type in cassava doesn't really provide a clean abstraction over headers + rows

This isn't probably the best time to get into it, but have you looked at [sv](http://hackage.haskell.org/package/sv) yet? It was just announced. Maybe it's better.

PR #395, opened 2018-03-15
Provide specialized volume metric lookup for participant metrics

kanishka-azimi:
- moving towards each "category" having specific functions instead of generic system,
start by exposing a function that provides active metrics used for a volume's participants
- this function is in preparation for csv upload parsing that will compare defined metrics
against csv fields detected

(This is slightly speculative, but the code using it should be merged tomorrow. Will change workflow once we come up with a new approach.)

### DIFF
--- a/src/Databrary/Model/Category.hs
+++ b/src/Databrary/Model/Category.hs
@@ -5,6 +5,7 @@ module Databrary.Model.Category
   , getCategory
   , getCategory'
   , categoryJSON
+  , participantCategory
   ) where
 
 import qualified Data.IntMap.Strict as IntMap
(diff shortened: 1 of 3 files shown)


src/Databrary/Model/Category.hs
```diff
@@ -72,3 +73,6 @@ categoryJSON :: JSON.ToObject o => Category -> JSON.Record (Id Category) o
 categoryJSON Category{..} = JSON.Record categoryId $
      "name" JSON..= categoryName
   <> "description" JSON..=? categoryDescription
+
+participantCategory :: Category
+participantCategory = head allCategories
```
kanishka-azimi: I guess I should use find and then fromMaybe... All options feel slightly messy.

kanishka-azimi:
Merging now, but will read feedback next week.

PR #396, opened 2018-03-15
Parsing a suggested mapping; stubs for routes

kanishka-azimi:
Unit tests still being added...

Continuing succession of merges resulting from recent work that builds up to the upload participant csv feature. Slightly speculative code, but awaiting synchronization with Bryan for how to avoid merging functions not used yet.

Feel free to review now, after unit tests have been added, or after merge. I will read and process comments whenever.

### DIFF
--- a/src/Databrary/Model/Ingest.hs
+++ b/src/Databrary/Model/Ingest.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE TemplateHaskell, QuasiQuotes, DataKinds #-}
+{-# LANGUAGE TemplateHaskell, QuasiQuotes, DataKinds, OverloadedStrings #-}
 module Databrary.Model.Ingest
   ( IngestKey
   , lookupIngestContainer
@@ -8,10 +8,15 @@ module Databrary.Model.Ingest
   , lookupIngestAsset
   , addIngestAsset
   , replaceSlotAsset
+  , HeaderMappingEntry(..)
   ) where
 
 import qualified Data.Csv as CSV
+import qualified Data.List as L
+import qualified Data.Map as Map
 import qualified Data.Text as T
+import Data.Text (Text)
+import Data.Map (Map)
 import Database.PostgreSQL.Typed.Query (pgSQL)
 import Database.PostgreSQL.Typed
 import Database.PostgreSQL.Typed.Query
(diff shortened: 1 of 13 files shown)


src/Databrary/Model/Ingest.hs
```diff
@@ -341,3 +352,57 @@ extractColumnsDistinctSampleJson maxSamples hdrs records =
...
+        if participantMetric `elem` volParticipantActiveMetrics
+        then 
+            case Map.lookup participantMetric requestedMapping of
+                Just csvField ->
+                    if (TE.encodeUtf8 csvField) `elem` colHdrs
+                    then pure (Just csvField)
+                    else fail ("unknown column (" ++ (show csvField) ++ ") for metric (" ++ (show . metricName) participantMetric)
+                Nothing -> fail ("missing expected participant metric:" ++ (show . metricName) participantMetric)
+        else
+            pure Nothing
```
kanishka-azimi: Current design typically expresses this as a "deform" form in the controller layer. Seems messy to have the controller be aware of such detailed information about the internals of participant field mapping.. Maybe the column exists, unique columns and metric is an active volume participant metric check can happen in Ingest controller, and the conversion from validated Map to ParticipantFieldMapping can happen here..

src/Databrary/Model/Metric.hs
```diff
@@ -701,6 +763,10 @@ getMetric (Id i) = IntMap.lookup (fromIntegral i) metricsById
 getMetric' :: Id Metric -> Metric
 getMetric' (Id i) = metricsById IntMap.! fromIntegral i
 
+lookupParticipantMetricBySymbolicName :: Text -> Maybe Metric
+lookupParticipantMetricBySymbolicName symbolicName =
+    find (\m -> (Data.Text.filter (/= ' ') . Data.Text.toLower . metricName) m == symbolicName) participantMetrics
+
```
kanishka-azimi: debated whether to only expose this or expose constants above. seem better to have compiler check for typos rather than use strings in more places.

kanishka-azimi:
Last PR for a day or so. Will add the type checker and then package up that with remaining code and some tests.

kanishka-azimi:
There is some cruft in validating the mapping provided still references real columns and metrics. Could possibly get around this by saving the Haskell values to disk and having client submit and id that references a validated mapping, at the price of making the API more stateful and obtuse

src/Databrary/Model/Metric.hs
chreekat: We can further improve this in the future by creating explicit Haskell-enum<--->Database-enum datatype mappings

PR #401, opened 2018-03-20
Participant Upload - Top level behavior of detect route; Full implementation of run upload route 

kanishka-azimi:
I don't really feel ready to handle feedback yet, but waiting any longer will only make it worse.

Today through Thursday is a good window where we won't have so much manual testing accumulated that I will hesitate to make major changes. I am pretty occupied with other tasks, but I can take short breaks to do small refactorings, especially if the code changing is under unit test coverage.


src/Databrary/Controller/Ingest.hs
```diff
@@ -69,15 +91,161 @@ postIngest = multipartAction $ action POST (pathId </< "ingest") $ \vi -> withAu
...
+                        -- TODO: 
+                        -- mOldVal <- getOldVal metric record
+                        -- action = maybe (Upsert val) (\o -> if o == val then Unchanged else Upsert val)
+                        let measureAction = Upsert val
+                        case measureAction of
+                            Upsert newVal ->
+                                void (changeRecordMeasure (Measure record met newVal))
+                            Unchanged ->
+                                pure ()
+            Nothing ->
```
chreekat: To me, `Nothing -> pure ()` usually indicates a missed opportunity to use a Functor instance somewhere.

src/Databrary/Model/Ingest.hs
```diff
@@ -348,12 +358,117 @@ replaceSlotAsset o n = do
...
+    mGestationalAge <- extractIfUsed2 pfmGestationalAge validateParticipantGestationalAge
+    mPregnancyTerm <- extractIfUsed2 pfmPregnancyTerm validateParticipantPregnancyTerm
+    mBirthWeight <- extractIfUsed2 pfmBirthWeight validateParticipantBirthWeight
+    mDisability <- extractIfUsed2 pfmDisability validateParticipantDisability
+    mLanguage <- extractIfUsed2 pfmLanguage validateParticipantLanguage
+    mCountry <- extractIfUsed2 pfmCountry validateParticipantCountry
+    mState <- extractIfUsed2 pfmState validateParticipantState
+    mSetting <- extractIfUsed2 pfmSetting validateParticipantSetting
+    pure
+        (ParticipantRecord
```
chreekat: These are all just `Maybe ByteString`? That seems really brittle. This, however, is the the part of Haskell records that is really tough: dealing with ambiguously-typed data. So maybe we just have to deal with it. Still, newtypes would help a lot, and tests are critical.

chreekat: It also makes me think ParticipantRecord should be a newtype over `Map ParticipantDatum ByteString`

src/Databrary/Model/Ingest.hs
```diff
@@ -348,12 +358,117 @@ replaceSlotAsset o n = do
...
+-- left if no match possible
+determineMapping :: [Metric] -> [Text] -> Either String ParticipantFieldMapping
+determineMapping participantActiveMetrics csvHeaders = do
+    (columnMatches :: [Text]) <- traverse (detectMetricMatch csvHeaders) participantActiveMetrics
+    let metricColumnMatches :: Map Metric Text
+        metricColumnMatches = (Map.fromList . zip participantActiveMetrics) columnMatches
+    -- TODO: sanity check -- all cols distinct
+    pure
+        (ParticipantFieldMapping
+            { pfmId = Map.lookup participantMetricId metricColumnMatches
```
chreekat: Same as above: maybe ParticipantFieldMapping should just *be* the Map. The keys can be easily converted to an enum rather than loose strings in this case, since it's just converting existing records (of the current ParticipantFieldMapping) to unary value constructors the hypothetical enum.

src/Databrary/Model/Metric.hs
```diff
@@ -770,6 +788,74 @@ lookupParticipantMetricBySymbolicName symbolicName =
 participantMetrics :: [Metric]
 participantMetrics = filter ((== participantCategory) . metricCategory) allMetrics
 
+validateParticipantId :: BS.ByteString -> Maybe BS.ByteString
```
chreekat: Related to the question of `ParticipantRecord`-as-Map, these functions might be a lot more palatable if defined like

```haskell
validateDatum :: ParticipantDatum -> BS.ByteString -> Maybe BS.ByteString
```

...giving one top-level function instead of dozens.

src/Databrary/Model/Record.hs
```diff
@@ -46,6 +49,22 @@ lookupVolumeRecord :: MonadDB c m => Volume -> Id Record -> m (Maybe Record)
...
+lookupVolumeParticipant vol idMeasureVal = do
+    allVolumeRecords <- lookupVolumeRecords vol
+    let mIdMatch = find matchesId allVolumeRecords
+    -- check record type is participant, if not, then error
+    pure mIdMatch
+  where
+    idMetric :: Metric
+    idMetric = getMetric' (Id 1)
+    matchesId :: Record -> Bool
+    matchesId r =
```
chreekat:
```
matchesId r = isJust $ find ...
```

test/Databrary/Model/IngestTest.hs
```diff
@@ -2,26 +2,70 @@
...
 
 import Databrary.Model.Metric
+import Databrary.Model.Record
 import Databrary.Model.Ingest
 import Databrary.Model.Metric.TypesTest
 
 tests :: TestTree
 tests = testGroup "Databrary.Model.Ingest"
     [ testCase "parseParticipantFieldMapping-1"
-        (parseParticipantFieldMapping [] [] (Map.fromList []) @?= Right emptyParticipantFieldMapping)
```
chreekat: I'd be down for -X OverloadedLists in (and only in) tests

src/Databrary/Controller/Ingest.hs
```diff
@@ -69,15 +91,161 @@ postIngest = multipartAction $ action POST (pathId </< "ingest") $ \vi -> withAu
...
+        mGestationalAge = getFieldVal prdGestationalAge "gestationalage"
+        mPregnancyTerm = getFieldVal prdPregnancyTerm "pregnancyterm"
+        mBirthWeight = getFieldVal prdBirthWeight "birthweight"
+        mDisability = getFieldVal prdDisability "disability"
+        mLanguage = getFieldVal prdLanguage "language"
+        mCountry = getFieldVal prdCountry "country"
+        mState = getFieldVal prdState "state"
+        mSetting = getFieldVal prdSetting "setting"
+    -- print ("save measure id:", mId)
+    changeRecordMeasureIfUsed recordStatus (Just (idVal, idMetric))
```
chreekat: This duplication of `changeRecordMeasureIfUsed` seems error-prone. Might be better as a `traverse` at least.

kanishka-azimi:
> The one thing I really have a stake in is the idea of making ParticipantRecord (and ParticipantFieldMapping, I think) actually be Map under a newtype.

We should talk about this one in person. It's definitely the hardest choice for me to make. The record ensures that each field is handled, so that once we have gone from json to the record, no other functions need to check for whether the key is in the map. I will explain in person as well, as I considered a few options before the version you see now.

src/Databrary/Model/Metric.hs
kanishka-azimi: There is obviously also the option of a type class as well... Mainly what happens in each validation will pretty unique, so my best hope is to build up combinators like not empty to use within the validations. Joy's hope is that eventually we expose user defined validations (DOB between 1970 and 1975)...

src/Databrary/Controller/Ingest.hs
kanishka-azimi: Just coding fast, definitely can condense. I actually changed this already on my latest branch for another bug fix.

src/Databrary/Controller/Ingest.hs
```diff
@@ -69,15 +91,161 @@ postIngest = multipartAction $ action POST (pathId </< "ingest") $ \vi -> withAu
...
 detectParticipantCSV = action POST (pathJSON >/> pathId </< "detectParticipantCSV") $ \vi -> withAuth $ do
-    pure $ okResponse [] ("" :: String)
+    v <- getVolume PermissionEDIT vi
+    csvFileInfo <-
+      -- TODO: is Nothing okay here?
+      runFormFiles [("file", maxWidelyAcceptableHttpBodyFileSize)] (Nothing :: Maybe (RequestContext -> FormHtml TL.Text)) $ do
+          csrfForm
+          fileInfo :: (FileInfo TL.Text) <- "file" .:> deform
+          return fileInfo
+    liftIO (print ("after extract form"))
```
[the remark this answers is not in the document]
kanishka-azimi: yes. just going sloppy, can probably switch to the logger

chreekat:
I looked into it a little bit and saw how big of a project it would be to make them maps, going straight down to the database schema, really. It shouldn't block this PR, though I still think it has a lot of value.

kanishka-azimi:
It's not that much work, the code is fresh in my head and there are tests for most of the impacted areas. Mostly comes down to a brief talk between us to decide whether the Map is flexible enough and/or too flexible or just right. ParticipantRecord and ParticipantFieldMapping are new, in the past, similar code used lists of Metrics (https://github.com/databrary/databrary/blob/master/src/Databrary/Ingest/JSON.hs#L242)

kanishka-azimi:
I merged in some recent changes I had to make for some bug fixes. That branch addressed one of the comments you made, most of the others are still any issue.

kanishka-azimi:
I'll start addressing the smaller ones tomorrow in between tasks.

test/Databrary/Model/IngestTest.hs
kanishka-azimi: Never heard of it, sounds cool and very helpful

src/Databrary/Model/Ingest.hs
kanishka-azimi: I want participantrecord to eventually have real types for each field. I was starting with bytestring because it was a smaller intermediate step from the existing implementation (Measure).

src/Data/Csv/Contrib.hs
```diff
@@ -30,12 +49,43 @@ extractColumnsDistinctSample maxSamples hdrs records =
...
+
+runCsvParser
+    :: (ATTO.Parser (Csv.Header, Vector Csv.NamedRecord) -> BS.ByteString -> Either String (Csv.Header, Vector Csv.NamedRecord))
+    -> BS.ByteString
+    -> Either String (Csv.Header, Vector Csv.NamedRecord)
+runCsvParser parse contents =
+    parse (Csv.csvWithHeader Csv.defaultDecodeOptions) (repairCarriageReturnOnly contents)
+
+-- | only fix newlines for bizarre macOS endings that use \r instead of \r\n
+repairCarriageReturnOnly :: BS.ByteString -> BS.ByteString
```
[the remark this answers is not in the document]
kanishka-azimi: The python library provides this as a parameter, but ignores the parameter currently. I can't remember if sv scored higher on this. Cassava's goal is RFC compliant and I believe macOS Excel CSV export is not compliant with it on newlines.

src/Databrary/Controller/Ingest.hs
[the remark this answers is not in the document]
kanishka-azimi: All of this logging was more a reaction to not having tests early on and having to troubleshoot reproduced bugs identified during manual testing. If I had the unit tests earlier, I wouldn't have even needed these.

kanishka-azimi:
Decided to switch everything to Maps. ideally at least Haskell enum as well, but might run out of time for that now.

kanishka-azimi:
unit test: all fields except id are "", expected action?

chreekat:
> unit test: all fields except id are "", expected action?

Sounds like a business question.

kanishka-azimi:
> Sounds like a business question.
There is an existing precedent as users already enter individual participants into the application. I just need to match the existing behavior. I wanted to record the expectation as a unit test here (https://github.com/databrary/databrary/blob/kanishka-dbd-845-a1/test/Databrary/Controller/IngestTest.hs#L32), mostly likely only one upsert for id.

chreekat:
I think I need more context to understand the question. :)

chreekat:
If it were implemented as a Map, would it suffice to check
```
fromList x @?= [(Id, "val")]
```
?

chreekat:
Oops I meant toList, naturally.

kanishka-azimi:
The case match is a hack around there not being an Eq instance derived for the types in question.

kanishka-azimi:
The original comment was cryptic as it was mostly a reminder to myself. Context:

Participant records are stored as triples of parent record id, field name/type, and field value. There is a way for users to currently enter in a participant record where they type a value for a field, then tab to the next field. This entry sends a value to the server as the user pushes tab or leaves focus. There is some convention implemented where either "delete" is sent or empty string is sent, which triggers deletion of a single record-key-value triple. I need to document exactly how that behaves and use the "deletion" logic when I get a CSV batch upload and any field in a row has an empty string.

kanishka-azimi:
I am starting the refactor to use Maps on branch tomorrow.

test/Databrary/Model/IngestTest.hs
kanishka-azimi: Most of the maps are gone for now, will add extension when there is boilerplate like this in future tests

kanishka-azimi:
@chreekat Thanks a lot for detailed feedback earlier.

src/Databrary/Controller/Ingest.hs
```diff
@@ -4,27 +4,53 @@ module Databrary.Controller.Ingest
   , postIngest
   , detectParticipantCSV
   , runParticipantUpload
-  -- for tests
+  -- for tests
```
chreekat:
```
mv .git/hooks/pre-commit.sample .git/hooks/pre-commit
```
^ Catches whitespace errors

src/Databrary/Controller/Ingest.hs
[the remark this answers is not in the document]
chreekat: I think these are worth cleaning up. Doesn't have to be done now, but we can open an issue or something.

src/Databrary/Model/Record/Types.hs
```diff
@@ -45,6 +48,27 @@ data Record = Record
 instance Kinded Record where
   kindOf _ = "record"
 
+-- keep value as a pair of parsed and raw value until we come up with a clever way to dispatch on type
+data ParticipantRecord = -- each field can be nothing (not used) or just value, where value can be empty
```
chreekat: This could be a tristate datatype for clarity. 

```haskell
data ParticipantRecordField = Unused | Empty | Field ByteString MeasureDatum
```
sure, we could get all uniplatey and phantom typey eventually, but I think creating a new, local, specailized type should be the standard action in these situations.

chreekat: Or moving to Map,
```haskell
data ParticipantRecordField = Empty | ParticipantRecordField MeasureDatum
newtype ParticipantRecord =
    ParticipantRecord { unParticipantRecord :: Map ByteString ParticipantRecordField }
```

PR #424, opened 2018-04-09
Copied them in and made them compile

chreekat:


chreekat:
No need for in depth review, just fyi. Ignore and merge blindly if it suits you :man_dancing:

test/Databrary/StringUtilTest.hs
```diff
@@ -0,0 +1,20 @@
...
+   ( tests )
+where
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+import Databrary.StringUtil
+
+tests :: TestTree
+tests = testGroup "Databrary.Model.EnumStringUtil"
```
chreekat: Oops. Here's a reason using tasty-discover might be nicer

chreekat: (I thought about it, and decided I did *not* want to rewrite everything just now)

test/Databrary/Service/PasswdTest.hs
```diff
@@ -0,0 +1,17 @@
...
+import Test.Tasty.HUnit
+
+import Databrary.Service.Passwd
+
+tests :: TestTree
+tests = testGroup "Databrary.Service.Passwd"
+  [ testCase "passwdCheck-1" (do
+       p <- initPasswd
+       mErr <- passwdCheck "pass" "user" "john" p
+       (mErr @?= Just "it is too short"))
```
kanishka-azimi: future note: make the error returned a type, so we can match on type instead of string

src/Databrary/StringUtil.hs
```diff
@@ -0,0 +1,26 @@
...
+  | otherwise = c:fromCamel' s
+
+toCamel :: String -> String
+toCamel "" = ""
+toCamel (c:s) = toUpper c:toCamel' s
+
+toCamel' :: String -> String
+toCamel' "" = ""
+toCamel' ('_':s) = toCamel s
+toCamel' (c:s) = c:toCamel' s
```
kanishka-azimi: I was hoping we could delete this eventually and replace with third party implementation. We'll see

databrary.cabal
```diff
@@ -540,10 +539,13 @@ test-suite databrary-test
...
+    , attoparsec
     , base
     , bytestring
     , containers
+    , http-types
     , postgresql-typed
+    , range-set-list
     , tasty
     , tasty-expected-failure
     , tasty-hunit
```
kanishka-azimi: test-suite module depends

chreekat: good catch

kanishka-azimi:
```
src/Databrary/Model/Citation.hs:24:57: fatal:
    cannot find object file for module ‘Databrary.Model.EnumStringUtil’
    while linking an interpreted expression
```

run ghci or cabal build ; fix up the minor compiler errors like above when you get a chance.

chreekat:
Oops, I expected this but couldn't get it to happen even after a cabal
clean. I suspect it's a new-build bug, caching something that shouldn't be
cached.

kanishka-azimi:
Thanks for doing this. One less thing to remember to do. I am also to the point where I run unit tests before each commit. I have been planning on switching to a nicer organization with more nested test groups, with names like "examples", "simple", and "advanced" under each function.

chreekat:
Actually, do *you* need a cabal clean? src/Databrary/Model/Citation.hs:24:57 is deep in some TH. I don't have any reference to "EnumStringUtil" in my repo anymore.

chreekat:
Ghci works for me too

kanishka-azimi:
Hm. Home machine similarly starts without a problem. Should probably re-enable `rm -rf dist` or add `cabal clean` inside of ghci-databrary, so that other developers don't experience this when they change branches. My work laptop still has the issue, so I will try different changes to ghci-databrary tomorrow.

kanishka-azimi:
`nix-shell --run "cabal -j2 new-test" ` - that causes the build failure for me. 

```
$ cd ~/dev/databrary && nix-shell --run "cabal -j2 new-test" 
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.
Resolving dependencies...
In order, the following will be built (use -v for more details):
 - databrary-1 (lib) (file src/Databrary/Controller/Ingest.hs changed)
 - databrary-1 (test:databrary-test) (configuration changed)
Preprocessing library for databrary-1..
Building library for databrary-1..
[ 70 of 292] Compiling Databrary.Model.Citation ( src/Databrary/Model/Citation.hs, /home/kanishka/dev/databrary/dist-newstyle/build/x86_64-linux/ghc-8.0.2/databrary-1/build/Databrary/Model/Citation.o ) [Databrary.Model.Audit changed]
src/Databrary/Model/Citation.hs:24:57: fatal:
    cannot find object file for module ‘Databrary.Model.EnumStringUtil’
    while linking an interpreted expression
cabal: Failed to build databrary-1 (which is required by test:databrary-test
from databrary-1).
```

kanishka-azimi:
```rm -rf dist-newstyle``` was the only thing that worked . there is now `cabal new-clean` yet. That triggered a pretty expensive rebuild, so I'm hesitant to force deletion of dist-newstyle every time tests run. I guess we can just note it as a repair step in the how to for running unit tests.

PR #426, opened 2018-04-10
Use tasty-discover

chreekat:
## Drawbacks
The one real drawback here is all the extraneous deps that
tasty-discover has. I could probably whip up a simple thing in bash that
does the same thing with zero deps.

## Using the repl

The repl must be restarted when new modules are added. However, new tests (and changed tests) are picked up by a :reload. Get to the tests with:
```
$ nix-shell --run 'cabal repl test:discovered'
<snip>
ghci> :load test/Discovered.hs
<snip>
ghci> main
```


kanishka-azimi:
Merge whenever you want.

kanishka-azimi:
Note to self, adding a test module:
1. create test module w/failing test
2. add test module to .cabal
3. (repl restart, if using repl)
4. run tests, observe failure

PR #430, opened 2018-04-12
Wall-clean: close enough

chreekat:
There are still a lot of type defaulting warnings due to OverloadedStrings + literal strings in debug messages, which should probably be removed, anyway. I just set -Wno-type-defaults.

There are also still the avFormat.c warnings, which would take no small amount of time to figure out the fix for.


chreekat:
I also think I might have missed some warnings in cached modules. Can fix them up in later passes.

kanishka-azimi:
About 3 of the removed variable names I had added as minimal documentation, short of a haddock, but not a big deal. Go for the merge whenever.

PR #431, opened 2018-04-13
Mechanism to notify users when conversion is down

kanishka-azimi:
- HPC (conversion cluster) goes down for scheduled maintenance for a full business day occasionally
- This change provides a mechanism for us to display a message to the user and allow the system to finish startup even when transcoding is unavailable
- Remove obsolete build flag to enable "sandbox" mode, since it closely overlaps with modified code.

Enabled by a flag config in databrary.conf combined with restarting the service. See example.conf.


src/Databrary/Controller/Root.hs
```diff
@@ -48,7 +48,9 @@ viewRootHandler api _ = -- TOOD: ensure GET
 
 viewConstantsHandler :: [(BS.ByteString, BS.ByteString)] -> Action
 viewConstantsHandler _ = -- TODO: ensure GET
-  withoutAuth $ return $ okResponse [] $ JSON.objectEncoding constantsJSON
+  withoutAuth $ do
+    (serv :: Service) <- peek
+    return $ okResponse [] $ JSON.objectEncoding (constantsJSON (serviceNotificationBar serv))
```
kanishka-azimi: This notification bar value could drift from the statically generated constants.js
Should probably not expose notificationbar in this view of constants.

PR #433, opened 2018-04-16
AssetSegment.Types test

chreekat:
Blindly moving in test code from the library module.

### DIFF
--- a/test/Databrary/Model/AssetSegment/TypesTest.hs
+++ b/test/Databrary/Model/AssetSegment/TypesTest.hs
@@ -1,4 +1,4 @@
-module Databrary.Model.AssetSegment.TypesTest where
+module Databrary.Model.AssetSegment.TypesTest (test_all) where
 
 import Test.Tasty
 import Test.Tasty.HUnit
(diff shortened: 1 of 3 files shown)


chreekat:
"On a partially-shared volume, excerpts should only be shown to the databrary community ("shared") but not to public"

kanishka-azimi:
> "On a partially-shared volume, excerpts should only be shown to the databrary community ("shared") but not to public"

Starting to remember better now. (This PR captures many relevant changes - https://github.com/databrary/databrary/pull/276/files?diff=unified)

Folders (sessions), video files (assets), and video file highlights (excerpts) each have a release level stored with the item. Sometimes, a finer grained item can default to its containing item (an asset might have an undefined release, so it falls back to the session's release level). Example values for a release level are: private, shared, and public. The release defines the maximum allowed sharing for a given item.

A volume also contains a high level toggle. In the past, this toggle could either be public or private. A public volume shared the maximum possible and a private volume hid everything. We introduced a partially shared volume to which shared some high level details about a volume, acting as a preview.

A user's effective access level for a given volume is computed as a function of what they can directly access and what their authorizing parties can access. Once that effective access value is computed, it is used to determine whether the user is viewing a volume privately, as part of the databrary community, or as the general public.

When a user is viewing an asset on a given volume, one compares the user's computed access level on the volume with the release level on the asset.

For partially shared volumes, all data like sessions and assets are meant to be hidden completely, very similar to how private volumes behave. This ignores the release level stored with a given asset or session. Currently. I believe the data is only hidden from the public, but not from the databrary community. I have yet to find time to implement the last change which will ensure data is only shared privately with people who have direct read access on a partially shared volume. Finally, within all these rules, there is one exception. When an excerpt has its release level set to public, then only the excerpt (and not the full asset) is shared with everyone, for partially shared volumes.

These tests address changes made while enabling excerpts on partially shared volumes to be visible by the public.

The key parts of the implementation are:
- downloadAssetSegment - retrieve a segment (time subset) of an asset, serve the underlying video content
- getAssetSegment - retrieve an asset segment, determining permission of requesting user in the process
- checkPermission2 - considering the volume in which the asset segment resides (getAssetSegmentVolumePermission2), ensure the user has at least read ("public") permission on the volume
- checkDataPermission3 - perform a deeper check to ensure the user's access level on the volume is high enough to view the given asset segment, consulting the asset segment's release level
- dataPermission3 - guts of checkDataPermission3, with the special case for partially shared volumes which have permission level "(PermissionPUBLIC, PublicRestricted)"
- getAssetSegmentRelease2 - considering session, the asset, and any excerpt that properly contains the segment being retrieved, figure out what is the most specific available release value present, applying a default if no release values have been explicitly entered by the user

EffectiveRelease is meant to capture the difference between the release level when a user has restricted versus full access to a volume, based on a combination of the volume's sharing level and the item's release level. This matters in particular when considering excerpts.

The actual instances of asset, segment, etc, read from the database rely subtlety on this: 
https://github.com/databrary/databrary/blob/master/src/Databrary/Model/AssetSegment/SQL.hs#L48

kanishka-azimi:
I went through a few iterations of reading, understanding, and modifying the code until it started all make sense together.

testdataPermission3 - good warm up, without lots of dependencies to understand
getAssetReleaseMaybe, getSlotReleaseMabye - simpler accessor functions that build up to the coalescing functions under test

When writing the tests, I wanted assurance that I understood most of the important possible combinations of values present in the fields of AssetSegment, and that getAssetSegmentRelease2 computed what I desired for each of those combinations.

test/Databrary/Model/AssetSegment/TypesTest.hs
```diff
@@ -0,0 +1,190 @@
...
+assetReleaseForPartialAndFullShared =
+    EffectiveRelease ReleasePUBLIC ReleasePUBLIC
+privateReleaseIgnoreSharing = EffectiveRelease ReleasePRIVATE ReleasePRIVATE
+excerptReleaseForPartialAndFullShared =
+    EffectiveRelease ReleasePUBLIC ReleasePUBLIC
+excerptReleaseForPartialAndFullShared' =
+    EffectiveRelease ReleaseSHARED ReleaseSHARED
+excerptReleaseForPartialAndFullShared'' =
+    EffectiveRelease ReleasePRIVATE ReleasePRIVATE
+
```
kanishka-azimi: The names of the variables were my attempts to denote the significance of each test case. They probably will translate into nested test group and test case names. The number of possibilities are confusing enough that both unit and property tests might help here.

chreekat:
This is a fantastic write-up, thanks! I'll synthesize it into better tests (and maybe something on Confluence)

kanishka-azimi:
My ideal would be to test something like:
```
getAssetSegment' :: Permission -> Bool -> AssetSegment -> AccessResult AssetSegment

data AccessResult = Hidden | InsufficientPermission | CanAccess AssetSegment
```
(or the simpler getAsset') which is some new pure helper function that getAssetSegment uses to perform the permission access logic, probably existing in the model layer. I tried to extract some pure logic, but I didn't bring it to a conclusion.

PR #437, opened 2018-04-18
AssetSlot.Types test

chreekat:
Copied in, similar to AssetSegment.Types' test.

@kanishka-azimi can you do a similar write up for this? Or is it all part of the same thing?


kanishka-azimi:
> Or is it all part of the same thing?

Yes

PR #439, opened 2018-04-18
foldIdentity simple example tests

kanishka-azimi:


test/Databrary/Model/Identity/TypesTest.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE OverloadedStrings, ScopedTypeVariables #-}
+module Databrary.Model.Identity.TypesTest where
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+import Databrary.Model.Identity.Types
+
```
kanishka-azimi: @chreekat Low value test, but it does demonstrate some general organization I have grown to like. Feel free to object or propose alternatives.

test/Databrary/Model/Identity/TypesTest.hs
```diff
@@ -0,0 +1,20 @@
...
+test_foldIdentity :: TestTree
+test_foldIdentity = testGroup "foldIdentity"
+    [ testCase "example" 
+          (runFoldIdentity NotIdentified @?= True)
+    , testCase "typical" (do
+          runFoldIdentity PreIdentified @?= True
+          runFoldIdentity (ReIdentified undefined) @?= True
+          runFoldIdentity (Identified undefined) @?= False)
+    ]
```
kanishka-azimi: "example", "typical", and "edge case" a blocks of tests, each within a test case. I find a test case for every example a little too noisy and too much boilerplate. I am essentially using testCase as a testGroup with unlabeled individual tests. Usually the tasty error gives the line number, so you can still pinpoint the failure.

test/Databrary/Model/Identity/TypesTest.hs
```diff
@@ -0,0 +1,20 @@
...
+
+runFoldIdentity :: Identity -> Bool
+runFoldIdentity = foldIdentity True (const False)
```
kanishka-azimi: Some simple utility function wrapping the function under test, to make the test cases more concise.

test/Databrary/Model/Identity/TypesTest.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE OverloadedStrings, ScopedTypeVariables #-}
+module Databrary.Model.Identity.TypesTest where
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+import Databrary.Model.Identity.Types
+
+test_foldIdentity :: TestTree
+test_foldIdentity = testGroup "foldIdentity"
```
kanishka-azimi: is there a way that discovered can add the "foldIdentity" test group for me?

chreekat: Did it not?

chreekat: Oh wait, I get what you mean.

Not within a single file, no. You'd have to create a new file.

If it's important to you, I'm almost positive I could write a 5-10 line replacement of tasty discover in bash that would make it a snap.

kanishka-azimi: I meant any way that I wouldn't have to type  `testGroup "foldIdentity" [ ... ]` , with some magic making the testgroup for me.... from the example, it looks like I can just provide the list. I'll try it later.

test/Databrary/Model/Identity/TypesTest.hs
chreekat: I don't feel strongly about it, we can see what develops. I will say this : while I believe tests should serve as documentation, I don't mean to take it too literally. Ideally one atomic semantic change should break exactly one test. (Never mind that semantic atomicity is probably very hard to pin down; it's just an ideal.)

test/Databrary/Model/Identity/TypesTest.hs
chreekat: To me, this indicates the function under test is weak in some way: not general enough, or maybe not specific enough.

For what it's worth, after reading these tests I still have zero clue what the function's semantics are, even if I could guess at its implementation. To me that indicates that the tests are basically reimplementing it rather than analyzing it.

test/Databrary/Model/Identity/TypesTest.hs
chreekat: Oh, so a test_foo could have type [Test tree]? That would be slick! Suddenly my shell script idea doesn't sound so great...

kanishka-azimi: yeah, see latest merge, it works...

test/Databrary/Model/Identity/TypesTest.hs
kanishka-azimi: > try to never use Bool unless the thing in question is actually a Boolean value

we happens in test vegas, stays in test vegas...... corrected in follow-up commit

test/Databrary/Model/Identity/TypesTest.hs
kanishka-azimi: I think you are spot on regarding good tests. I once spent almost 1.5 times as much effort defining properties for some geometric algorithms as I did implementing the algorithms.

But there is also a bootstrapping problem here. We don't have documentation or properties/tests for most of the implementation, so expected behavior is either obvious or we have to learn it through experimentation and reverse engineering the produced data from user's so far. I think it's important to get the documenting and learning phase moving, so that we can get to the point where we both have enough generators (and fake instances) to quickly express a property, and also to start recording what we think is the underlying behavior and why. For me, this phase involves writing as many short "reimplementing function" tests as possible, and correcting those tests as we realize our simple assumptions are wrong. Then, I would start circling back and replacing those useless tests with the real test suite and documentation. At that point, I would aim to generalize into useful generators and property checkers expressing relationships between functions, complemented by one very simple unit test example (essentially a doctest) per function, and write anything that can't be inferred from above in haddocks. The desire to write a longer haddock can be a smell that we need more properties.

I am not prescribing that for you, just letting you know why I am writing what feel like either obvious or useless tests. I am also happy to adopt other bootstrapping approaches if you have something in mind. I imagine the best way to bootstrap straight into good property checkers is to start by writing confluence pages about each function, and then extracting logical properties from the long English narratives.

BTW, fake might fit within my vision for the example ("doctests") tests.

chreekat: You are absolutely correct that we are bootstrapping, and that this isn't really about tests, it's about making sense of the domain.

Can we perhaps distinguish between "bootstrap" tests and real tests somehow? Maybe no action required at the moment since we only have the one type.

PR #449, opened 2018-04-20
Duplicate ghci-databrary installed script into standalone script, mak…

kanishka-azimi:
…e more robust

- standalone script is a little easier to hack than inline string in default.nix
- still need to remove from default.nix when dev documentation is updated
- ensure dependencies are available in databrary-dev derivation
- create config directory on startup if it doesn't exist
----
This was done at this moment so I could reuse the script from the int test directory.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,7 +14,7 @@ let
   nixpkgs = reflex-platform.nixpkgs;
   inherit (nixpkgs.lib) id;
   nodePackages = import ./node-default.nix { pkgs = nixpkgs; };
-  inherit (nixpkgs) fetchFromGitHub writeScriptBin cpio wget;
+  inherit (nixpkgs) fetchFromGitHub writeScriptBin cpio wget git;
   # nixpkgs functions used to regulate Haskell overrides
   inherit (nixpkgs.haskell.lib)
     dontCheck overrideCabal doJailbreak doCoverage doHaddock dontHaddock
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -87,7 +87,13 @@ let
...
       # cabal override to enable ghcid (GHCi daemon) development tool
       databrary-dev = overrideCabal databrary (drv: {
-        libraryHaskellDepends = (drv.libraryHaskellDepends or []) ++ (with self; [ghcid cabal-install ghciDatabrary]);
+        libraryHaskellDepends =
+	  (drv.libraryHaskellDepends or [])
+	   ++
+	     (with self;
+	       [ghcid cabal-install ghciDatabrary
+	       # for ghci-databrary script
+	       wget cpio nodePackages.shell.nodeDependencies git]);
```
kanishka-azimi: I'm not sure if installing git into nix store will interfere with any settings in system installed nix for ubuntu/centos users.

ghci-databrary.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+set -e
+
+# this is replacing default.nix installation of this script into nix store, remove that soon
+# default.nix should still reference the nix store items this depends on!!
+git_root_dir=$(git rev-parse --show-toplevel)
```
kanishka-azimi: Added this so that this script could be used from any directory where /src and databrary.cabal are in the current directory. Will use this in integration tests to startup from runtime-dirs/databrary-inttest

kanishka-azimi:
cc: @chreekat

kanishka-azimi:
note to self, use something like the following in startup script for int test:
```
nix-shell
$ cd runtime-dirs/databrary-inttest
$ ../../ghci-databrary.sh
> ...
```

PR #452, opened 2018-04-20
First integration test, disorganized

kanishka-azimi:
~~Let's criticize this and come up with a few short cleanup steps.~~ Below is one test. We'd have some wrapper loop or even simple a long shell script to invoke each test.

```
# some TBD cleanup/reset
$ runtime-dirs/databrary-inttest/inttest-start-db.sh   #maybe create a simple process group somehow?
$ runtime-dirs/databrary-inttest/inttest-start-backend.sh
$ cd runtime-dirs/databrary-inttest/controllers/root/robotstxt/ && ./runtest.sh
# shutdown db, backend
# return exit code
```

Note: fetchresult.hs is meant to output the result in some standard format soon, so that we could interchangeably use curl or wget if desired for some tests.

~~I'll merge this after that initial cleanup.~~

~~After that, we can use this as something to critique into what we actually want. Do we prefer:~~

~~* like this? ("unix philosophy-ish" - this code is dedicated to Dan, a die-harder unix philosophy man)~~
   ~~* if so, move more source files into pkgs/databrary-inttest?~~
~~* bats?~~
~~* use more shell instead of haskell?~~
~~* combine fetch and check scripts?~~
~~* go all out haskell with database, app startup, and full tasty test suite~~
   ~~* if so, compiled?~~
~~* systemd instead of just shell?~~

### DIFF
--- a/runtime-dirs/databrary-inttest/controllers/root/robotstxt/runtest.sh
+++ b/runtime-dirs/databrary-inttest/controllers/root/robotstxt/runtest.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+git_root_dir=$(git rev-parse --show-toplevel)
+nix-shell --attr databrary-dev.env $git_root_dir/default.nix --command "runghc fetchresult.hs | runghc checkresult.hs"
--- a/databrary.cabal
+++ b/databrary.cabal
@@ -595,4 +595,26 @@ test-suite discovered
     , Data.Csv.ContribTest
     , Data.RangeSet.ParseTest
 
+test-suite inttest-placeholder
+  type: exitcode-stdio-1.0
+  main-is: Main.hs
+  hs-source-dirs: pkgs/databrary-inttest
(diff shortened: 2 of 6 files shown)


runtime-dirs/databrary-inttest/controllers/root/robotstxt/runtest.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
kanishka-azimi: I know this file can be vastly improved... I don't have a clear vision yet for exactly when to enter the nix-shell and where to place each script and what working directories to use.

chreekat:
As mentioned in other (closed) PR, I'm going to delay having an opinion until I've seen a lot more unit tests.

kanishka-azimi:
Okay... This is the integration test driver. Just let me know any time you want to weigh your opinion in integration test framework. I am happy to gut this at any point.

PR #454, opened 2018-04-23
Expand model.transcode to underlying queries

kanishka-azimi:
Testing manually currently before merge.

First example of queries that replace something that used the ORM before.


kanishka-azimi:
all but findTranscode and addTranscode (update path) were tested. rare for the untested paths to be used, and very tiny chance that purely syntactic change would introduce different behavior.

PR #455, opened 2018-04-23
Half of Container model queries expanded

kanishka-azimi:
Manually testing in progress.


kanishka-azimi:
Tested all impacted routes manually.

PR #481, opened 2018-04-26
Login path, initial unit tests

kanishka-azimi:


kanishka-azimi:
cc: @JeffSpies , if you are curious.

JeffSpies:
Good start. Did I hear you say before that we already have data fixtures when hitting the db in  (future) tests?

kanishka-azimi:
>  we already have data fixtures when hitting the db

I have been adding object level data generators. These could be used eventually in tests that drive create or update methods.

The way the database gets created for development instance has some limited test data bootstrapped in. We could easily add more test data there until we come up with a more robust mechanism. 

(https://github.com/databrary/databrary/blob/master/initializeDB.sh#L9 uses each migration in the schema directory, including the test data here  https://github.com/databrary/databrary/blob/master/schema/0.sql#L958 )

JeffSpies:
Data: 👍 

How is 0.sql generated? Dump from the db post migrations? Generated by an app?

kanishka-azimi:
I don't know how 0.sql was generated, it was long before I started. Most likely a snapshot from a development db instance, manually edited to include test data.

There used to a script for generating database dumps, which hasn't been actively maintained for a while.

At this point, whenever we decide to collapse down the current migrations, we can create a new script for generating a fresh 0.sql. For a little while longer, we can add test data in place over 0.sql, since the file is only actively used for development and testing databases currently.

PR #509, opened 2018-05-01
do not merge: used for diff review

chreekat:


src/Databrary/Model/Identity/Types.hs
```diff
@@ -16,15 +16,20 @@ import Databrary.Model.Permission.Types
 import Databrary.Model.Party.Types
 import Databrary.Model.Token.Types
 
+-- | ????
 data Identity
   = PreIdentified
```
chreekat: It can't be for background jobs, since BackgroundContext uses NotIdentified.

PR #519, opened 2018-05-03
Very detailed documenting of cascading Party permission logic on load of party

kanishka-azimi:
This mostly makes sense to me, but will probably only be easy to understand a few months from now if we have a wall of example style unit tests, enumerating all the typical possibilities.

The only thing slightly cumbersome about writing example tests or factoring out logic for testing this function is the recursive reference to `p`.

I would be shocked if there was no subtle bug within this coalescing logic, but probably an unimportant bug.

cc: @chreekat

### DIFF
--- a/src/Databrary/Model/Party/SQL.hs
+++ b/src/Databrary/Model/Party/SQL.hs
@@ -56,6 +56,7 @@ selectPermissionParty = selectJoin 'makeParty
   ]
 
 -- | Build an account or party, based on calling context.
+-- Compute permission and access by coalescing authorization granted directly and generally
 permissionParty
   :: Has (Id Party) a
   => (Permission -> Maybe Access -> a) -- ^ Partially applied makeParty, ready to build full account or party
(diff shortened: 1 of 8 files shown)


src/Databrary/Model/Party/SQL.hs
```diff
@@ -42,40 +42,72 @@ selectPartyRow = selectColumns 'PartyRow "party" ["id", "name", "prename", "orci
...
+            -- for Identified, the access is the levels granted to the user on the databrary site via it's parent
+            $ accessSite' generalSiteAccessForViewer
     mAccessDeduced :: Maybe Access
     mAccessDeduced
-      | foldIdentity False (((view p :: Id Party) ==) . view) ident = Just maxBound
-      | identityAdmin ident = Just $ maybe id (<>) access1 $ view ident
-      | otherwise = access1
+      -- if the viewing identity is Identified, and the viewer is the same as the party being retrieved,
+      --  then allow unbounded permission on the retrieved party (self)
+      | extractFromIdentifiedSessOrDefault
```
chreekat: What's going on with whitespace here? Tabs? (Starting line 95)

chreekat:
I'm a little concerned about rewriting the logic at this stage, though I appreciate that the new stuff is slightly easier to follow.

src/Databrary/Model/Party/SQL.hs
kanishka-azimi: I wanted to use a let in, but since it was a guard, I kept it to one line. The arguments don't fit in one line easily.

PR #528, opened 2018-05-10
Kanishka test 0510

kanishka-azimi:


kanishka-azimi:
Tested each impacted function manually.

PR #529, opened 2018-05-10
Kanishka expand party

kanishka-azimi:


kanishka-azimi:
Tested each query manually

PR #533, opened 2018-05-14
Change lookupAuthorize to conditionally retrieve inactive entries

kanishka-azimi:
There is a bug where an authorized investigator attempts to delete an expired authorization, but the deletion doesn't actually take effect. The problem was the deletion was relying on a lookupAuthorize to find the record to delete, but lookupAuthorize was implemented to only retrieve active authorizations.

Now, lookupAuthorize will conditionally include all authorizations when searching for a match, and deleteAuthorize route uses this enhanced functionality.

### DIFF
--- a/src/Databrary/Controller/Authorize.hs
+++ b/src/Databrary/Controller/Authorize.hs
@@ -157,7 +157,7 @@ postAuthorize = action POST (pathAPI </>> pathPartyTarget </> pathAuthorizeTarge
 -- Also, build an authorization request or present the current authorization value.
 findOrMakeRequest :: (MonadDB c m) => Party -> Party -> m (Maybe Authorize, Authorize)
 findOrMakeRequest child parent = do
-  c <- lookupAuthorize child parent
+  c <- lookupAuthorize ForPartyViewer child parent -- TODO: should this conditionally use ForPartyAdmin based on permissionParty?
   pure (c, mkAuthorizeRequest child parent `fromMaybe` c)
 
 -- | If present, delete either a prior request for authorization. The authorization to delete can be specified
@@ -170,7 +170,7 @@ deleteAuthorize = action DELETE (pathAPI </>> pathPartyTarget </> pathAuthorizeT
       mAuthorizeTargetParty <- lookupParty oi
       maybeAction mAuthorizeTargetParty
   let (child, parent) = if apply then (p, o) else (o, p)
-  mAuth <- lookupAuthorize child parent
+  mAuth <- lookupAuthorize ForPartyAdmin child parent
   removeAuthorizeNotify mAuth
   case api of
     JSON -> return $ okResponse [] $ JSON.pairs $ "party" JSON..=: partyJSON o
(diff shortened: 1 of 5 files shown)


src/Databrary/Model/Authorize.hs
```diff
@@ -58,16 +62,24 @@ lookupAuthorizedChildren parent perm = do
     (\p -> $(selectQuery (selectAuthorizeChild 'parent 'ident) "$WHERE (expires IS NULL OR expires > CURRENT_TIMESTAMP) AND site >= ${p} AND member >= ${p} AND (site <> 'NONE' OR member <> 'NONE')"))
     perm
```
kanishka-azimi: Below is the only real change, along with usage of this at the call site

kanishka-azimi:
Was referring to this a bit while I was implementing - https://github.com/databrary/databrary-dev-learning/blob/master/dumped-splices/Model/Authorize.dump-splices

src/Databrary/Model/Authorize.hs
```diff
@@ -58,16 +62,24 @@ lookupAuthorizedChildren parent perm = do
...
 
+data AuthorizeFilter = ForPartyAdmin | ForPartyViewer deriving (Eq, Show)
```
chreekat: These constructor names introduce a level of indirection. One must understand that PartyAdmins get all authorizations, while PartyViewers only get active auths. I suggest

```haskell
data AuthorizationFilter = AllAuthorizations | ActiveAuthorizations
```

src/Databrary/View/Party.hs
```diff
@@ -43,7 +43,7 @@ htmlPartyView :: Party -> RequestContext -> H.Html
 htmlPartyView p@Party{ partyRow = pr@PartyRow{..}, ..} req = htmlTemplate req Nothing $ \js -> do
   H.div H.! H.customAttribute "typeof" "person" $ do
     H.h1 H.! H.customAttribute "property" "name" $ H.text $ partyName pr
-    when (view p >= PermissionEDIT) $
+    when (partyPermission >= PermissionEDIT) $
```
chreekat: :tada:

src/Databrary/Model/Party/Types.hs
```diff
@@ -65,8 +65,8 @@ instance Has (Id Party) Account where
 instance Has Access Party where
   view Party{ partyAccess = Just a } = a
   view _ = mempty
-instance Has Permission Party where
-  view = partyPermission
+-- instance Has Permission Party where
```
chreekat: Delete

src/Databrary/Model/Authorize.hs
```diff
@@ -58,16 +62,24 @@ lookupAuthorizedChildren parent perm = do
...
+
 -- | Attempt to find an authorization request or grant from the child party to the granting parent party.
 -- Exclude expired authorizations.
```
chreekat: Needs haddock update

src/Databrary/Controller/Authorize.hs
```diff
@@ -170,7 +170,7 @@ deleteAuthorize = action DELETE (pathAPI </>> pathPartyTarget </> pathAuthorizeT
       mAuthorizeTargetParty <- lookupParty oi
       maybeAction mAuthorizeTargetParty
   let (child, parent) = if apply then (p, o) else (o, p)
-  mAuth <- lookupAuthorize child parent
+  mAuth <- lookupAuthorize ForPartyAdmin child parent
   removeAuthorizeNotify mAuth
```
chreekat: If you're motivated, `removeAuthorizeNotify` should be refactored from `:: Maybe Authorize -> k` to `:: Authorize -> k`, with fmaps applied where needed at call sites. Obviuosly this is not critical to this PR.

kanishka-azimi: Agreed, I noticed that before as well. I don't understand all the notification behavior in updateAuthorize, so I won't change it this time around.

PR #534, opened 2018-05-14
Build out more examples of creating authorized parties

kanishka-azimi:


kanishka-azimi:
Should expand testharness to provide more helpers for creating parties, contexts, and authorizations.

PR #551, opened 2018-05-25
Kanishka assetslot expand

kanishka-azimi:
Needs manual test for each affected query before merge.


kanishka-azimi:
Tested each query manually.

PR #568, opened 2018-05-29
Kanishka cleanup unit

kanishka-azimi:
Start moving the rushed tests for authorize and related over to a centralized area. Eventually, I imagine having about five of these BDD style modules, group by either an aspect (e.g. authorization) or a domain (e.g. records, assets). 

@chreekat 

We were both withholding judgment on how we wanted these unit/component tests to be organized. ModelTest is close to how I want these tests to look like. It would be nice to move some of the helper functions over to the library. Some of the helper functions might become unnecessary if we find ways to abstract the parameters or return values further.

I also envision having a doctest style example for each function in the corresponding Model.MODULETest module.

For the BDD tasty tests, I would like keep the suite pretty small and focused. I would also want Katalon, Runscope/bats+curl testing a single example, whereas many combinations might be tested at this layer or lower.

### DIFF
--- a/test/TestHarness.hs
+++ b/test/TestHarness.hs
@@ -1,6 +1,8 @@
 module TestHarness
     (
       TestContext ( .. )
+    , mkDbContext
+    , runContextReaderT
     , withinTestTransaction
     , stepsWithTransaction
     , connectTestDb
@@ -9,6 +11,7 @@ module TestHarness
     , addAuthorizedInstitution
     , mkInstitution -- TODO: stop exporting
     , mkAccount -- TODO: stop exporting
+    , mkAccountSimple -- TODO: stop exporting
     , addAuthorizedInvestigator
     , addAffiliate
     , lookupSiteAuthNoIdent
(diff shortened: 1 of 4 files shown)


test/TestHarness.hs
```diff
@@ -181,6 +194,16 @@ addAuthorizedInvestigator adminCtxt lastName firstName email instParty = do
...
 
+addAuthorizedInvestigatorWithInstitution :: DBConn -> BS.ByteString -> T.Text -> BS.ByteString -> IO (Account, TestContext)
+addAuthorizedInvestigatorWithInstitution cn adminEmail instName aiEmail = do
+    ctxt <- makeSuperAdminContext cn adminEmail
+    instParty <- addAuthorizedInstitution ctxt instName
+    aiAcct <- addAuthorizedInvestigator ctxt "Last" "First" aiEmail instParty
+    let ctxtNoIdent = ctxt { ctxIdentity = IdentityNotNeeded, ctxPartyId = Id (-1), ctxSiteAuth = view IdentityNotNeeded }
+    Just aiAuth <- runReaderT (lookupSiteAuthByEmail False aiEmail) ctxtNoIdent
+    let aiCtxt = switchIdentity ctxt aiAuth False
+    pure (aiAcct, aiCtxt)
```
kanishka-azimi: This is one key helper function that might be need almost everywhere. Will move to TestContext soon... I'm starting to want a few different general helper modules. TestContext, Factories, ...

PR #569, opened 2018-05-30
Kanishka expand 0529

kanishka-azimi:


kanishka-azimi:
Manually tested usage of each query.

PR #570, opened 2018-05-30
Kanishka cleanup 0530

kanishka-azimi:
Almost done, will merge soon.


kanishka-azimi:
Will move the "TypesTest" modules to one shared "generators" module soon, as was requested a while back.

PR #573, opened 2018-06-01
Begin Servant, Part II

chreekat:
This is another PR that is mostly just pulling apart the ball of wax. Comments, formatting, and surgical refactors. No change of behavior, and no specific changes to the implementation of individual routes. *Routing* has had a few changes, however. In particular, the order in which route parsing happens has changed. The new order is: Servant -> ActionRoute -> Wai.Routes.

* [x] Smoketest run


chreekat:
I haven't run the smoketest yet

src/Databrary/Action/Response.hs
```diff
@@ -110,18 +131,26 @@ emptyResponse s h = response s h BS.empty
...
 okResponse = response ok200
 
+-- | A wrapper for the short-circuiting machinery (see 'result').
 newtype Result = Result { resultResponse :: Response } deriving (Typeable)
 instance Show Result where
   showsPrec p (Result r) = showParen (p > 10)
     $ showString "Result " . showsPrec 11 (responseStatus r)
 instance Exception Result
 
+-- | Short circuit immediately, returning the given Reponse.
```
chreekat: Typo

src/Databrary/Action/Servant.hs
```diff
@@ -0,0 +1,58 @@
...
+    ) where
+
+import Servant
+
+import Databrary.API
+import Databrary.Action (ActionRouteApp (..))
+
+-- | Create a server to serve the ServantAPI.
+apiServer :: ActionRouteApp -> Server ServantAPI
+apiServer (ActionRouteApp app) = preferAngularServer :<|> Tagged app
```
chreekat: Better name: servantServer

src/Databrary/Model/Identity.hs
```diff
@@ -31,8 +31,15 @@ determineIdentity :: (MonadHas Secret c m, MonadHasRequest c m, MonadDB c m) =>
...
-maybeIdentity u i = extractFromIdentifiedSessOrDefault u i =<< peek
+-- | Takes default action and a monadic function. If the Identity within the
+-- monadic context is 'Identified', apply the function to the 'Session' held
+-- within. Otherwise, run the default action.
+maybeIdentity
+    :: (MonadHasIdentity c m)
+    => m a -- ^ Default action
+    -> (Session -> m a) -- ^ Monadic function
+    -> m a -- ^ Result
+maybeIdentity z f = extractFromIdentifiedSessOrDefault z f =<< peek
```
chreekat: extractFromIdentifiedSessOrDefault should be renamed maybeIdentity, and the existing maybeIdentity should be deleted.

kanishka-azimi: I can't remember, but I think i made it wordy intentionally, because the original name was too concise.

src/Databrary/Action/Run.hs
```diff
@@ -122,27 +120,54 @@ actionApp service (Action needsAuth act) waiReq waiSend
...
+    , ctxSec :: Secret
+    , ctxConn :: DBConn
+    }
+
+instance Has Wai.Request IdContext where view = ctxReq
+instance Has Secret IdContext where view = ctxSec
+instance Has DBConn IdContext where view = ctxConn
+
+-- | Look up the user's identity (or don't)
+fetchIdent
```
kanishka-azimi: I sometimes use "resolve" instead of "fetch" for functions like this. Although determine identity and resolve identity would have some ambigious overlap in their names.

src/Databrary/Action/Run.hs
```diff
@@ -66,53 +66,51 @@ data NeedsAuthentication = NeedsAuthentication | DoesntNeedAuthentication
...
-                            (RequestContext ctx waiReq identity)
-              return (identity, waiResponse)
+actionApp service (Action needsAuth act) waiReq waiSend =
+    let
+        isdb = isDatabraryClient waiReq
+        authenticatedAct :: ActionContextM (Identity, Response)
+        authenticatedAct = do
+            sec <- peek
+            conn <- peek
+            identity <- fetchIdent sec conn waiReq needsAuth
```
kanishka-azimi: Applicative could bring the two peeks into this action line

src/Databrary/API.hs
```diff
@@ -12,23 +12,24 @@
...
+-- The fallback to OG routing
+type ActionRouteAPI = Raw
 
--- | Boilerplate. This will go away when Servant adopts type application.
-servantAPI :: Proxy ServantAPI
-servantAPI = Proxy
+-- | Routes that fire up the Angular SPA if possible.
+type PreferAngularAPI
+    = Header "User-Agent" Text :> QueryParam "js" Int :> EmptyAPI
+        -- Up next:
```
kanishka-azimi: Consider going dirtier for a little longer and copying and pasting the header and query param check to each route, until you have repeated the pattern for about 4 or so routes, and then decide whether that alone is the abstraction or if more should be added

src/Databrary/API.hs
```diff
@@ -12,23 +12,24 @@
...
--- Up next:
---
---    = "user" :> "login" :> Get '[HTML] Html
---    :<|> LegacyAPI
+type ServantAPI
+    = PreferAngularAPI
+    :<|> ActionRouteAPI
+
+-- The fallback to OG routing
+type ActionRouteAPI = Raw
```
kanishka-azimi: Once you start matching urls, I would also add a failure match for each url, to capture the error earlier. For example, if the url matches, but other parts don't match, it might be better to have servant stop any fallback attempts there...  Or you might intentionally want the opposite behavior, where servant matches a url for html, but let it fallback to old routes for json req on same url, during transition.

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
...
-            then
-                -- Our hatch to the WaiRouteApp is Servant!
-                servantApp waiRouteApp req
-            else
-                actionApp routeContext (err (st,hdrs)) req
+actionRouteApp invMap (WaiRouteApp waiRouteApp) svc = ActionRouteApp
+    (\req resp ->
+        -- Use the original Action if it still exists
+        either
+            (waiRouteFallback req resp)
```
kanishka-azimi: It would help if you explicitly named the params here, as I would usually read this code in passing

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
...
+            (\act -> actionApp svc act req resp)
+            (Invertible.routeWai req invMap)
+    )
   where
+    waiRouteFallback req resp (st,hdrs)
+        -- Currently, this might be only possible error result?
```
kanishka-azimi: I think I guessed this from reading the web-inv-route source code

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
 maybeAction Nothing = result =<< peeks notFoundResponse
```
kanishka-azimi: ActionRouteApp is a little misleading, as it has WaiRouteApp inside of it... Consider something like:
InvertiblePlusWaiRouteApp and WaiRouteApp

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
...
 
 newtype WaiRouteApp = WaiRouteApp Application
+newtype ActionRouteApp = ActionRouteApp Application
```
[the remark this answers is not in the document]
kanishka-azimi: Tagged might be a compromise between explicit types and less indirection.

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
...
+            (\act -> actionApp svc act req resp)
+            (Invertible.routeWai req invMap)
+    )
   where
+    waiRouteFallback req resp (st,hdrs)
+        -- Currently, this might be only possible error result?
+        | st == notFound404 = waiRouteApp req resp
+        -- Handle any other possible errors with the original app's error
+        -- handling.
+        | otherwise = actionApp svc (err (st,hdrs)) req resp
```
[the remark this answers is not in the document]
chreekat: I didn't look -- this was a straight refactor, keeping the original logic

src/Databrary/Action.hs
```diff
@@ -83,47 +83,38 @@ maybeAction (Just a) = return a
...
-    -- ^ The original route map. Now partially replaced by Servant and Wai
-    -- Routes
+    -- ^ The original route map. Now partially replaced by Wai Routes
     -> WaiRouteApp
     -- ^ The newer Wai Route-based Application
     -> Service
     -- ^ System capabilities
-    -> Wai.Application
+    -> ActionRouteApp
     -- ^ The actual web app
```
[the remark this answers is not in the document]
chreekat: I don't disagree. I do like maybe, but either seems a bit much.

src/Databrary/Action/Run.hs
chreekat: I have a habit of using 'fetch' for things that hit the database. What's your criteria for something to be called 'resolve'?

CONTRIBUTING.md
```diff
@@ -0,0 +1,8 @@
+If you have made architectural changes, you can run a quick smoketest of major
+features with the following steps:
+
+* Login (use user/pass test@databrary.org/unastan54)
+* Create volume
+* Add participant (try Karen Adolph)
```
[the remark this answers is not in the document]
chreekat: Nope, that's new to me. Maybe walk me through it

src/Databrary/Action/Run.hs
kanishka-azimi: Usually I use "resolve" when it is some wrapper function that ends up calling "get by id" at the model layer. Sometimes the resolve function is also in charge of producing the message in the Left when the id is invalid.

PR #595, opened 2018-06-20
Kanishka unit 0619

kanishka-azimi:
Exercise search from within tests

### DIFF
--- a/src/Databrary/Solr/Index.hs
+++ b/src/Databrary/Solr/Index.hs
@@ -24,6 +24,7 @@ import Network.HTTP.Types.Header (hContentType)
 import Control.Invert
 import Databrary.Ops
 import Databrary.Has
+import Databrary.HTTP.Client (HTTPClient)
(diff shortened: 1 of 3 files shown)


src/Databrary/Solr/Index.hs
```diff
@@ -184,7 +185,7 @@ joinContainers f cl@(c:cr) al@((a, SlotId ci s):ar)
   | containerId (containerRow c) == ci = f a (Slot c s) : joinContainers f cl ar
   | otherwise = joinContainers f cr al
 
-writeVolume :: (Volume, Maybe Citation) -> SolrM ()
+writeVolume :: (Volume, Maybe Citation) -> ReaderT BackgroundContext (InvertM BS.ByteString) ()
```
kanishka-azimi: Breaking this function and writeAllDocuments down to use constraints with much more narrow requirements than BackgroundContext would make test setup for solr indexing much simpler.

PR #597, opened 2018-06-22
Remove superfluous path hierarchy

chreekat:
* [x] make test
* [x] make db
* [x] make repl
* [ ] nix-build -A databrary

I didn't let nix-build run to completion yet, but it *started* just fine.


chreekat:
I'm going to try running the coverage-report test against it

chreekat:
I think I *won't* merge this. The code would probably work fine, but it's just too disruptive to git history. `git blame` would show my name for every single line of code! :D

chreekat:
Oh wait, no it doesn't. git is smarter than that.

chreekat:
fyi @myicicle : Haskell source files got moved. Now there's less typing and noise, at the expense of needing to use `git log --follow <file>` to see the complete history for a file.

PR #598, opened 2018-06-25
Kanishka minor 0625

kanishka-azimi:


src/Model/VolumeState/Types.hs
```diff
@@ -7,12 +7,13 @@ import qualified Data.Aeson as JSON
...
 import Model.Volume.Types
+import Model.Id.Types
 
 type VolumeStateKey = T.Text
 
 data VolumeState = VolumeState
   { volumeStateKey :: !VolumeStateKey
   , volumeStateValue :: !JSON.Value
   , volumeStatePublic :: !Bool
-  , stateVolume :: !Volume
```
kanishka-azimi: I'd like to do this more generally.

PR #600, opened 2018-06-26
Explicit types for post and put request bodies

kanishka-azimi:
These are the types I want Servant to parse into. Can switch from plain types to records at a later point. I have wanted to perform this small step for months.

### DIFF
--- a/src/Controller/Asset.hs
+++ b/src/Controller/Asset.hs
@@ -172,6 +172,15 @@ detectUpload u = do
   either deformError' (return . FileUpload u)
     =<< lift (probeFile (fileUploadName u) =<< peeks (fileUploadPath u))
 
+data ProcessAssetRequest =
+    ProcessAssetRequest
+        (Maybe (FileInfo TempFile))
+        (Maybe (Id Token))
+        (Maybe T.Text)
+        (Maybe Release)
+        (Maybe (Id Container))
+        ()
+
 processAsset :: AssetTarget -> Handler Response
 processAsset target = do
   let as@AssetSlot{ slotAsset = a, assetSlot = s } = case target of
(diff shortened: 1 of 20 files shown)


kanishka-azimi:
Eventually, there should be at least one generator in "model.factories" to loosely correspondence with each request type here.

kanishka-azimi:
@chreekat I am merging this. It's a superficial change and seems universal good and the PR has been open for a few days.

I am happy to rollback this change or modify it, if you do have feedback later.

src/Controller/Asset.hs
```diff
@@ -172,6 +172,15 @@ detectUpload u = do
...
     =<< lift (probeFile (fileUploadName u) =<< peeks (fileUploadPath u))
 
+data ProcessAssetRequest =
+    ProcessAssetRequest
+        (Maybe (FileInfo TempFile))
+        (Maybe (Id Token))
+        (Maybe T.Text)
+        (Maybe Release)
+        (Maybe (Id Container))
+        ()
```
chreekat: To verify, you're saying you'd like `processAsset` to have type `:: AssetTarget -> ProcessAssetRequest -> handler response` ?

kanishka-azimi:
> My only concern, as usual, is that now we have a bunch of "dead" code to keep track of. However, making an account of our API types is a huge win. Let's just do servant sooner rather than later!

Another two hours and we could probably delete the function that I partially replaced.

kanishka-azimi:
Actually, wrong PR, ignore last comment.

src/Controller/Asset.hs
kanishka-azimi: Haven't decided yet, but that sounds promising. processAsset itself is way too big.

An alternative would be to extend AssetTarget to include more data.

PR #601, opened 2018-06-27
For review only

kanishka-azimi:
Start outline of simpler json generation for asset

### DIFF
--- a/src/Controller/Asset.hs
+++ b/src/Controller/Asset.hs
@@ -109,6 +109,40 @@ assetJSONQuery :: AssetSlot -> JSON.Query -> Handler (JSON.Record (Id Asset) JSO
 assetJSONQuery o q = (assetSlotJSON False o `JSON.foldObjectIntoRec`) <$> JSON.jsonQuery (assetJSONField o) q
 -- public restricted should consult volume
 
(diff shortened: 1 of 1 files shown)


kanishka-azimi:
This is one place where I think Haskell is behind other languages in terms of ORM flexibility. These sort of features would be built into the database library instead of requiring each application like databrary to invent its own convention.

chreekat:
It looks like what you're doing is describing an intermediate datatype to clarify a parse step. That seems like a good idea, and one of the things Haskell is uniquely suited for. What feature are you lacking in particular?

kanishka-azimi:
Rails and other frameworks make it really easy to load only certain relations, without having to implement extra data types or functions. Here, AssetExtended is trying to capture the full universe of possible data I might want to join with an Asset and I will have to write lookupAssetExtended myself instead of gluing together ORM function calls.... I think reusable parts of query as opaleye/beam/squeal provide comes close, but there is still some more boilerplate than the rich ORMs could provide.

kanishka-azimi:
BTW, in the current system, AssetCore = AssetRow and AssetExtended = Asset + a lot more fields

PR #605, opened 2018-06-28
Toehold for running ingest, upload, notifications in test context

kanishka-azimi:
This isn't polished, but I can take any interim feedback as I get ready to make it polished in the coming days.

### DIFF
--- a/test/TestHarness.hs
+++ b/test/TestHarness.hs
@@ -9,6 +9,8 @@ module TestHarness
     , withAV
     , withTimestamp
     , withLogs
+    , mkLogsStub
+    , mkNotificationsStub
     , withInternalStateVal
     , mkDbContext
     , runContextReaderT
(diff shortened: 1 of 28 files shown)


test/TestHarness.hs
```diff
@@ -316,6 +333,14 @@ withLogs ctxt = do
     logs <- initLogs mempty
     pure (addCntxt ctxt (blankContext { ctxLogs = Just logs }))
 
+mkLogsStub :: IO Logs
+mkLogsStub = initLogs mempty
+
+mkNotificationsStub :: IO Notifications
+mkNotificationsStub = do
+    conf <- C.load "databrary.conf" -- alternatively, don't use config
+    initNotifications (conf C.! "notification")
```
kanishka-azimi: Turn this recurring pattern into a function ```mkStubWithConfig :: (SomeClass b) => String -> (b -> a) -> IO a```

PR #609, opened 2018-06-30
Site authorization level in find parties result

kanishka-azimi:
This finishes up adding each parties site authorization level in the result for find parties endpoint. It is based on the implementation of same feature for individual parties throught the "authorization" query paramter.


src/Model/Party.hs
```diff
@@ -347,7 +351,7 @@ addParty bp = do
                 vaffiliation_a6PKi
                 vurl_a6PKj)
            row)
-  pure ((\p -> Party p Nothing PermissionREAD Nothing) pRow)
+  pure ((\p -> Party p Nothing Nothing PermissionREAD Nothing) pRow)
```
kanishka-azimi: Turn this repeated default party creation into function

src/Model/Party.hs
```diff
@@ -998,16 +1002,24 @@ findParties pf = do
   rows <- dbQuery $ unsafeModifyQuery -- (selectQuery (selectParty 'ident) "")
     (mapQuery2
```
kanishka-azimi: Should the coalesce happen in the query or makeParty2?

src/Model/Party.hs
```diff
@@ -998,16 +1002,24 @@ findParties pf = do
...
-                "SELECT party.id,party.name,party.prename,party.orcid,party.affiliation,party.url,account.email FROM party LEFT JOIN account USING (id) "])
+             -- TODO: this duplicates logic in lookupAuthorization slightly
+            [Data.String.fromString 
+                "SELECT \
+                \  party.id,party.name,party.prename,party.orcid,party.affiliation,party.url,account.email \
+                \ ,COALESCE(av.site, 'NONE') \
+                \ FROM party \
+                \   LEFT JOIN account USING (id) \
+                \   LEFT JOIN authorize_view av \
+                \      ON party.id = av.child AND av.parent = 0 "])
```
kanishka-azimi: Should lookupparty also be be modified now, or wait until later?

src/Model/Party/Types.hs
```diff
@@ -86,23 +88,28 @@ instance Has Access SiteAuth where
 
 deriveLiftMany [''PartyRow, ''Party, ''Account]
 
+-- The values below assume a minimalist loading of each object, with no
+-- related objects loaded.
```
kanishka-azimi: Should these be loaded with siteaccess value here?

src/Model/Party/Types.hs
```diff
@@ -121,6 +128,7 @@ nobodySiteAuth = SiteAuth
         , partyURL = Nothing
         }
       , partyAccount = Nothing
+      , partySiteAccess = Nothing -- not loaded
```
kanishka-azimi: Prob fill this here

test/Model/Party/TypesTest.hs
```diff
@@ -1,45 +0,0 @@
...
-                      Just
-                          (Access { accessSite' = PermissionSHARED, accessMember' = PermissionSHARED })
-                }
-        a =
-            Account {
-                  accountEmail = "john.smith@nyu.edu"
-                , accountParty = p
-                }
-    in
-        p
```
kanishka-azimi: Replicate this in model.factories

src/Model/Party.hs
chreekat: While I haven't taken the time to look into this question, I suspect the answer is "Either is fine; let's worry about converting to beam first."

chreekat: Same goes for the preceding question.

src/Model/Party/SQL.hs
```diff
@@ -135,8 +141,8 @@ selectAuthParty ident = selectMap (`TH.AppE` TH.VarE ident) $ selectJoin 'permis
 
 -- | Used by 'makeUserAccount' and 'selectPermissionAccount'. This finishes building the circular Party and Account structure.
 makeAccount :: PartyRow -> (Party -> Account) -> Permission -> Maybe Access -> Account
-makeAccount pr ac perm ma = a where
-  a = ac $ Party pr (Just a) perm ma
+makeAccount pr ac perm ma = a where -- TODO: take site auth permission
```
chreekat: This TODO could use some expansion, assuming it still exists in this form.

src/Model/Party/Types.hs
```diff
@@ -41,6 +41,8 @@ data Party = Party
   { partyRow :: !PartyRow
   , partyAccount :: Maybe Account
   -- , partySiteAccess :: Access -- site-level access this party is granted under root (currently SiteAuth only)
+  , partySiteAccess :: !(Maybe Permission) -- ^ site-level data access this party is granted under root.
```
chreekat: I don't understand the phrase "under root". I also don't know what a "loaded-only" value might be.

src/Model/Party/Types.hs
chreekat: Is there any chance for a 

```
-- | <More detailed explanation of what it means for a value to be "loaded">
data Loaded a 
    = Loaded something_here? a
    | Unloaded a
```

test/Model/Party/TypesTest.hs
chreekat: I worry that little comments like this will get lost unless they are tracked somewhere else. I'm sort of just letting them slip past, because I usually don't get the feeling that they're important enough to hold on to (or if they are, you will manage it yourself). Should we maybe capture them somewhere that we can review every week or two? Some of them may indeed be good to have as open issues, providing toe holds for wannabe volunteers.

kanishka-azimi: I would write "example tests" rather than keep a backlog of these. The tests can start as outlines with only steps to start, posing the question that needs clarification.

chreekat: Sorry, I mean the github comments, like https://github.com/databrary/databrary/pull/609#discussion_r199312812

PR #617, opened 2018-07-02
Expand Citation Queries

kanishka-azimi:
https://github.com/databrary/databrary/pull/574 - old version

Finish citation; needs testing.


kanishka-azimi:
unit tests cover this function, passed.

PR #621, opened 2018-07-03
getVolume: Separate Model response from Controller response

chreekat:
Creates a more testable abstraction between Model and Controller for this one function. Others could follow (issues incoming).

### DIFF
--- a/src/Controller/Permission.hs
+++ b/src/Controller/Permission.hs
@@ -12,7 +12,7 @@ module Controller.Permission
 import Control.Monad (void, unless, liftM2)
 
 import Has (Has, view, peek, peeks)
-import Model.Permission
+import Model.Permission hiding (checkPermission)
 import Model.Release
 import Model.Party
 import Model.Identity
@@ -20,7 +20,8 @@ import HTTP.Request
 import Action
 
 -- logic inside of checkPermission and checkDataPermission should be inside of model layer
-checkPermission :: Has Permission a => Permission -> a -> Handler a  -- TODO: delete this
+-- TODO: use Model.checkPermission everywhere instead
+checkPermission :: Has Permission a => Permission -> a -> Handler a
 checkPermission requiredPermissionLevel objectWithCurrentUserPermLevel =
   checkPermission2 view requiredPermissionLevel objectWithCurrentUserPermLevel
 
--- a/src/Controller/Permission.hs
+++ b/src/Controller/Permission.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Controller.Permission
-  ( checkPermission
-  , checkPermission2
+  ( checkPermissionOld
+  , checkPermission
   , userCanReadData
   , authAccount
   , checkMemberADMIN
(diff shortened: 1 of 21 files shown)


src/Controller/Permission.hs
```diff
@@ -12,15 +12,16 @@ module Controller.Permission
...
 import Model.Release
 import Model.Party
 import Model.Identity
 import HTTP.Request
 import Action
 
 -- logic inside of checkPermission and checkDataPermission should be inside of model layer
-checkPermission :: Has Permission a => Permission -> a -> Handler a  -- TODO: delete this
+-- TODO: use Model.checkPermission everywhere instead
+checkPermission :: Has Permission a => Permission -> a -> Handler a
```
kanishka-azimi: I was attempting to phase out the use of `view`... maybe we can replace `view` with `viewPermission` and `HasPermission` later. No action needed currently

src/Controller/Volume.hs
```diff
@@ -78,12 +78,23 @@ import Controller.Web
...
+    -- ^ The volume, as requested (or a short-circuited error response)
 getVolume p i = do
-  mVol <- lookupVolume i
-  vol <- maybeAction mVol
-  checkPermission p vol
+  resp <- Model.getVolume p i
+  case resp of
+    LookupFailed -> result =<< peeks notFoundResponse
+    LookupDenied -> result =<< peeks forbiddenResponse
+    LookupFound v -> pure v
```
kanishka-azimi: get rid of `result` later on and scrap Handler here, no action needed now

src/Model/Permission.hs
```diff
@@ -69,3 +72,28 @@ accessJSON :: JSON.ToObject o => Access -> o
 accessJSON Access{..} =
      "site" JSON..= accessSite'
   <> "member" JSON..= accessMember'
+
+-- | Responses to 'checkPermission'
+data PermissionResponse a
+    = PermissionGranted a
+    -- ^ Whatever you wanted, you got it!
+    | PermissionDenied
```
kanishka-azimi: Almost feels like an instance of bool like, similar to how the string like works

src/Controller/Volume.hs
chreekat: Wouldn't that require duplicating this logic everywhere Controller.getVolume is currently used, though? imo it makes sense to have a wrapper function that generates HTTP responses if that's the desired behavior (and by "desired" I mean "existing")

src/Model/Volume.hs
```diff
@@ -47,6 +49,41 @@ import Model.VolumeAccess.Types (VolumeAccess(..))
...
+-- does NOT mean "does not exist". It means that 'lookupVolume' (for example)
+-- returned Nothing. This could mean either the id is a valid id, or the user
+-- doesn't have access to the volume.
+data LookupResult a
+    = LookupFailed
+    | LookupDenied
+    | LookupFound a
+
+-- | Get a 'Volume', if it is available to the given 'Identity' with the desired
+-- 'Permission'.
```
kanishka-azimi: Ctrl feels wordy

src/Model/Volume.hs
```diff
@@ -47,6 +49,41 @@ import Model.VolumeAccess.Types (VolumeAccess(..))
...
+    => Permission -- ^ Desired permission level
+    -> Id Volume -- ^ Id of Volume to get
+    -> ctrl (LookupResult Volume)
+getVolume reqestedPermission volId = do
+    mVol <- lookupVolume volId
+    pure $ case mVol of
+        Nothing -> LookupFailed
+        Just vol ->
+            let
+                requiredPermission =
```
kanishka-azimi: This is permission granted or authorized, not required

chreekat:
One thing that seems dicey is adding all these `import Model.Foo hiding (...)`. On the whole, however, I feel like it's drawing light to existing code smell, rather than creating new smell.

test/ModelTest.hs
```diff
@@ -173,9 +173,11 @@ test_7 = Test.stepsWithTransaction "test_7" $ \step cn2 -> do
...
     vol <- runReaderT (addVolumeSetPrivate aiAcct) aiCtxt
     step "Then the public can't view it"
-    -- Implementation of getVolume PUBLIC
-    mVolForAnon <- runWithNoIdent cn2 (lookupVolume ((volumeId . volumeRow) vol))
-    mVolForAnon @?= Nothing
+    mVol <-
+        runWithNoIdent
+            cn2
+            (getVolume PermissionREAD ((volumeId . volumeRow) vol))
+    mVol @?= LookupFailed
```
kanishka-azimi: I would leave these looser, as not lookupFound

src/Controller/Volume.hs
kanishka-azimi: Yeah..maybe this would envelope the entire response then

chreekat: Yeah, this might become more general given more example uses. `getErr :: Permission -> Permissioned a -> Handler a`

src/Model/Permission.hs
chreekat: `OverloadedBool` --> Paving the way for the inaugural [IOHCC](https://ioccc.org/)

src/Model/Volume.hs
chreekat: Yeah, it's not great. Was trying to find something between `Handler` and the nondescript `m` that indicates *Controller*. But maybe `m` is fine

src/Model/Volume.hs
chreekat: Oh, good catch! That would have definitely tripped me up reading my own code in the future, when I would have assumed I knew what I was talking about when I wrote this.

test/ModelTest.hs
chreekat: I like making the tests prickly... "fail fast"

PR #627, opened 2018-07-05
Use test storage fixture

chreekat:
## Description
Use a storage fixture that centralizes some concerns:
* Create a directory structure in temp space
* Clean up after use

## Motivation and Context
* Tests not running on NixOS, which is a bit of a stickler for these things

## How Has This Been Tested?
* The tests themselves are what changed

## Checklist:


test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
 mkRequest :: Wai.Request
 mkRequest = Wai.defaultRequest { Wai.requestHeaderHost = Just "invaliddomain.org" }
 
+-- | Create a temp dir and use it for all Storage fields
+withStorage2 :: ResourceT IO Storage
```
chreekat: I'm going to switch to just using `bracket`, by the way.

test/ModelTest.hs
```diff
@@ -317,45 +318,48 @@ test_12a = ignoreTest $ -- "Invalid cross-device link"
...
+            let aiCtxt4 = withTimestamp (UTCTime (fromGregorian 2017 5 6) (secondsToDiffTime 0)) aiCtxt3
+            aiCtxt5 <- withLogs aiCtxt4
+            let aiCtxt6 = withInternalStateVal ist aiCtxt5
+            foundAsset <- runReaderT
+                (do
+                    (a, _) <- Gen.sample (genCreateAssetAfterUpload vol)
+                    let a' = a { assetRow = (assetRow a) { assetRelease = Just ReleasePUBLIC, assetFormat = (fromJust . getFormatByExtension) "webm" } }
+                    rpath <- liftIO (BSC.pack <$> getDataFileName "test/data/small.webm") -- TODO: touch timestamp to trigger new file
+                    Right probe <- probeFile "small.webm" rpath
+                    savedAsset <- addAsset a' (Just rpath)
```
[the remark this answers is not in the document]
kanishka-azimi: cp the file into upload directory in this test. have rpath reference the upload directory path.

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+        tmpPath <- unRawFilePath tmp
+        createDirectoryIfMissing False tmpPath
+    -- @initTemp x = mkdir -p "${x}/participantUpload"@
+    initTemp t = do
+        let tempPath = addTrailingPathSeparator t
+        unTempPath <- unRawFilePath tempPath
+        createDirectoryIfMissing
+            False
+            (getStorageTempParticipantUpload' unTempPath)
+        pure tempPath
```
kanishka-azimi: If you can delete initStorage now, and rename this, it will make reading this diff way easier.

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+        }
+  where
+    -- Check the dir exists (2nd arg) and it's on the same device as the 1st
+    -- arg (if Just).
+    checkDirs :: Maybe DeviceID -> RawFilePath -> IO (Maybe DeviceID)
+    checkDirs dev f = do
+        s <- getFileStatus f
+        f' <- unRawFilePath f
+        unless (isDirectory s) $ ioError $ mkIOError
+            doesNotExistErrorType
```
kanishka-azimi: impressive reuse

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+            (Just f')
+        let d = deviceID s
+        unless (all (d ==) dev) $ ioError $ mkIOError
+            illegalOperationErrorType
+            "storage filesystem"
+            Nothing
+            (Just f')
+        return $ Just d
+    -- @initCache x = mkdir -p "${x}/tmp"@
+    initCache c = do
```
kanishka-azimi: type sig

test/ModelTest.hs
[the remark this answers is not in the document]
kanishka-azimi: genCreateAssetAfterUpload - this should probably be in charge of that task

src/Store/Service.hs
chreekat: +1 It shall not live beyond this PR

transctl.sh
```diff
@@ -47,7 +52,11 @@ if [[ -n $test ]] ; then
 			rsync -p "$cmd" "$host:$hcmd"
 		fi
 	else
-		test -d "$dir"
+		if ! test -d "$dir"
+                then
+                    warn "$0: $dir does not exist"
+                    false
+                fi
```
kanishka-azimi: does false where set the exit code for below?

test/ModelTest.hs
```diff
@@ -317,45 +321,49 @@ test_12a = ignoreTest $ -- "Invalid cross-device link"
 -- same as above, but for video file that undergoes conversion
 test_12b :: TestTree
 -- FIXME: ignored because it hangs indefinitely for bryan
-test_12b = ignoreTest $ localOption NoTimeout $ Test.stepsWithResourceAndTransaction "test_12b" $ \step ist cn2 -> do
+test_12b = localOption NoTimeout $ Test.stepsWithResourceAndTransaction "test_12b" $ \step ist cn2 -> do
```
kanishka-azimi: Timeout of 25 seconds or less... we can take some averages between machines

test/ModelTest.hs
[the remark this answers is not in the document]
kanishka-azimi: Same device for performance (and space?) optimization, as the files are large video files.

test/ModelTest.hs
chreekat: I ran it for 10 minutes and it still didn't finish.

transctl.sh
chreekat: Yes. I don't like this style, but I didn't want to start rewriting this file just now.

chreekat:
I won't have time to finish this this week. It should be ok to leave it open though, yeah?

kanishka-azimi:
Sure.

test/ModelTest.hs
kanishka-azimi: Yes, see other notes about methods for detecting when conversion is complete. Running on nix, its encountering an infinite loop. If you look at the log file for the conversion, you will see some sort of error like ffmpeg not found.

chreekat: Where can I find these notes and log file?

kanishka-azimi: Can't remember if I noted it in a PR or in Slack. The note was something like "There are multiple ways for the test to detect when conversion has completed. I am considering moving towards starting up a single route warp server instead of tailing the log file."

The log file is referenced in the detect function - `"trans/" ++ show idVal ++ ".log"`

src/Store/Transcoder.hs
```diff
@@ -18,6 +22,36 @@ runTranscoder :: Transcoder -> [String] -> IO (ExitCode, String, String)
 runTranscoder (Transcoder cmd arg) args =
   readProcessWithExitCode cmd (arg ++ args) ""
 
+data TranscodeConfig = TranscodeConfig
+    { transcodeHost :: Maybe String
+    , transcodeDir :: Maybe String
+    , transcodeMount :: Maybe String
+    }
```
kanishka-azimi: later on, can turn this into a few constructors w/o Maybe's. Something like: `TranscodeDisabled | LocalTranscode String | RemoteTranscode String String String`

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+        let tmp = c </> "tmp"
+        tmpPath <- unRawFilePath tmp
+        createDirectoryIfMissing False tmpPath
+    -- @initTemp x = mkdir -p "${x}/participantUpload"@
+    initTemp t = do
+        let tempPath = addTrailingPathSeparator t
```
kanishka-azimi: Probably do this step before the function

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+    -> IO Storage
+    -- ^ The 'Storage' resource (presuming no exceptions were thrown).
+initStorage2 (Left e) _ = return $ error $ "Storage unavailable: " ++ e
+initStorage2 (Right StorageLocationConfig {..}) initTc = do
+    temp <- maybe (rawFilePath =<< getTemporaryDirectory) pure storageLocTemp
+    foldM_
+        checkDirs
+        Nothing
+        ([storageLocMaster, temp, storageLocUpload]
+        ++ catMaybes [storageLocStage]
```
kanishka-azimi: maybe

src/Store/Service.hs
```diff
@@ -10,13 +14,107 @@ import System.Directory (getTemporaryDirectory, createDirectoryIfMissing)
...
+    checkDirs dev f = do
+        s <- getFileStatus f
+        f' <- unRawFilePath f
+        unless (isDirectory s) $ ioError $ mkIOError
+            doesNotExistErrorType
+            "storage directory"
+            Nothing
+            (Just f')
+        let d = deviceID s
+        unless (all (d ==) dev) $ ioError $ mkIOError
```
kanishka-azimi: `unless (maybe True (d ==) mDev)` - I know you are just moving the code, but if you are interested.

test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
...
+    -- XDG_RUNTIME_DIR, TMPDIR, or "/tmp", whichever is available first
+    userTmpDir <- liftIO
+        (getEnvDefault "XDG_RUNTIME_DIR" =<< getEnvDefault "TMPDIR" "/tmp")
+    -- Explicitly drop the release key, since the ResourceT in which this runs
+    -- will always be short-lived.
+    (_, dir) <- allocate
+        (mkdtemp (userTmpDir </> "databrary-storage-fixture-"))
+        removeDirectory
+    -- initStorage and initTranscoder expect these to exist.
+    (liftIO  . mapM_ (flip createDirectory 0o755 . (dir </>)))
```
kanishka-azimi: maybe without flip?

test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
...
+        ["tmp", "stage", "upload", "trans", "cache"]
+    let tc = initTranscoder2 TranscodeConfig
```
kanishka-azimi: tc = localTransocding

test/ModelTest.hs
```diff
@@ -317,45 +321,49 @@ test_12a = ignoreTest $ -- "Invalid cross-device link"
...
-                  collectTranscode t 0 Nothing ""
-                  Just t2 <- lookupTranscode (transcodeId trans)
-                  pure (transcodeAsset t2))
-                  -- lookup assetslot or container w/contents
-        (aiCtxt6 { ctxSecret = Just (Secret "abc"), ctxRequest = Just mkRequest }) -- use mkRequest to override default domain
-    step "Then one can retrieve the resulting asset"
-    (assetRelease . assetRow) foundAsset @?= Just ReleasePUBLIC
-    (assetName . assetRow) foundAsset @?= Just "small.webm"
-    (assetDuration . assetRow) foundAsset @?= Just (Offset 5.62)
+    runResourceT $ do
```
kanishka-azimi: Might be shorter tests that use storage, or we can create one, so you don't have to reason about all of this at the same time as building the abstraction.

chreekat: Not a bad idea. I want to focus on getting the storage part squared away. My trouble with this test is now the following, which seems unrelated:

```
228 35
[swscaler @ 0x15bae00] deprecated pixel format used, make sure you did set range correctly
252 35.mp4
curl: (6) Could not resolve host: invaliddomain.org
```

chreekat: Sorry, that's not the "trouble"; that's the new log message. (The previous one was about fdk_aac being unavailable, which I fixed.) The trouble is that the test still hangs forever.

kanishka-azimi: That's a good sign. It means it completed and attempted the curl call back. Somehow the grep isn't recognizing the last line

chreekat:
I merged in origin/master to resolve the conflict, literally just as an experiment...

test/ModelTest.hs
kanishka-azimi: Definitely add the timeout now. It's designed to run infinitely on failure.

test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
 mkRequest :: Wai.Request
 mkRequest = Wai.defaultRequest { Wai.requestHeaderHost = Just "invaliddomain.org" }
 
+-- | Create a temp dir and use it for all Storage fields
```
kanishka-azimi: Take an argument indicating the base directory to copy initial contents for each subdirectory. For now, it can be the git root directory. Later, it can be use a special directory.

test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
...
+withStorage2 = do
+    -- XDG_RUNTIME_DIR, TMPDIR, or "/tmp", whichever is available first
+    userTmpDir <- liftIO
+        (getEnvDefault "XDG_RUNTIME_DIR" =<< getEnvDefault "TMPDIR" "/tmp")
+    -- Explicitly drop the release key, since the ResourceT in which this runs
+    -- will always be short-lived.
+    (_, dir) <- allocate
+        (mkdtemp (userTmpDir </> "databrary-storage-fixture-"))
+        removeDirectory
+    -- initStorage and initTranscoder expect these to exist.
```
kanishka-azimi: Expose everything from here down as a function. It will be useful for the fixture generation later.

kanishka-azimi:
Ensure file exists before grep, in detectjobdone

databrary.nix
```diff
@@ -101,6 +104,9 @@ in
     jdk
     coreutils # cat, cut, md5sum
   ];
+  testPkgconfigDepends = [
+    ffmpeg
```
kanishka-azimi: Turn on the store.av test now, if it's still off?

kanishka-azimi: Should probably add ffmpeg as tool depend, since transcode assumes ffmpeg is on the path.

Could add each she'll command used by transctl and transcode as tool depends as well

chreekat: ffmpeg was already added (it caused a failure on nixos). The latter is probably good hygiene, but beyond the 90% mark we're hitting for now

test/TestHarness.hs
```diff
@@ -282,6 +286,36 @@ instance Has Access TestContext where
 mkRequest :: Wai.Request
 mkRequest = Wai.defaultRequest { Wai.requestHeaderHost = Just "invaliddomain.org" }
 
+-- | Create a temp dir and use it for all Storage fields
+withStorage2 :: ResourceT IO Storage
+withStorage2 = do
+    -- XDG_RUNTIME_DIR, TMPDIR, or "/tmp", whichever is available first
+    userTmpDir <- liftIO
+        (getEnvDefault "XDG_RUNTIME_DIR" =<< getEnvDefault "TMPDIR" "/tmp")
```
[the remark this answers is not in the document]
chreekat: Would be good to confirm it exists on OSX first

test/TestHarness.hs
chreekat: Would take longer than 5 minutes... punting for now

chreekat:
Lots of followup, but I'm merging this as a Minimum Viable Change. Some followup is tracked by deprecations. Others will hopefully become apparent. Some docs on usage are necessary.

test/TestHarness.hs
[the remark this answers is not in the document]
kanishka-azimi: not on os x, only tmpdir

PR #633, opened 2018-07-06
Test to cover editing, removing a record on a volume

kanishka-azimi:
## Description

Test editing field data and removing a record altogether.

## Motivation and Context

increase test coverage

## How Has This Been Tested?

yes

## Types of changes

## Checklist:

- [x] I am sick of writing content for this PR, let's get this over with.

Note: Records can be created: 1. directly, 2. through participant upload, 3. through ingest.

### DIFF
--- a/test/Model/Factories.hs
+++ b/test/Model/Factories.hs
@@ -391,10 +391,7 @@ genGenderMeasure :: Gen (Metric, BS.ByteString)
 genGenderMeasure =
     pure (participantMetricGender, "Male")
 
--- TODO: genCreateMeasures -- will ensure each measure metric is distinct
--- genParticipantMetricValue :: Gen (Metric, BS.ByteString)
--- genParticipantMetricValue =
---    Gen.choice [genBirthdateMeasure, genGenderMeasure]
+-- TODO: genCreateMeasures :: Gen [(Metric, BS.ByteString)] -- will ensure each measure metric is distinct
 
 genCreateGenderMeasure :: Gen Measure
 genCreateGenderMeasure = do
--- a/test/Model/Factories.hs
+++ b/test/Model/Factories.hs
@@ -391,17 +391,10 @@ genGenderMeasure :: Gen (Metric, BS.ByteString)
 genGenderMeasure =
     pure (participantMetricGender, "Male")
 
-genParticipantMetricValue :: Gen (Metric, BS.ByteString)
-genParticipantMetricValue =
-    Gen.choice [genBirthdateMeasure, genGenderMeasure]
-
-genCreateMeasure :: Gen Measure -- TODO: don't use this
-genCreateMeasure = do
-    (mtrc, val) <- genParticipantMetricValue
-    Measure
-        <$> (pure . error) "measure record not set yet"
-        <*> pure mtrc
-        <*> pure val
+-- TODO: genCreateMeasures -- will ensure each measure metric is distinct
+-- genParticipantMetricValue :: Gen (Metric, BS.ByteString)
+-- genParticipantMetricValue =
+--    Gen.choice [genBirthdateMeasure, genGenderMeasure]
 
 genCreateGenderMeasure :: Gen Measure
 genCreateGenderMeasure = do
--- a/test/Model/Factories.hs
+++ b/test/Model/Factories.hs
@@ -395,13 +395,30 @@ genParticipantMetricValue :: Gen (Metric, BS.ByteString)
 genParticipantMetricValue =
     Gen.choice [genBirthdateMeasure, genGenderMeasure]
 
-genCreateMeasure :: Gen Measure
+genCreateMeasure :: Gen Measure -- TODO: don't use this
 genCreateMeasure = do
     (mtrc, val) <- genParticipantMetricValue
     Measure
         <$> (pure . error) "measure record not set yet"
         <*> pure mtrc
         <*> pure val
+
+genCreateGenderMeasure :: Gen Measure
+genCreateGenderMeasure = do
+    (mtrc, val) <- genGenderMeasure
+    Measure
+        <$> (pure . error) "measure record not set yet"
+        <*> pure mtrc
+        <*> pure val
+
+genCreateBirthdateMeasure :: Gen Measure
+genCreateBirthdateMeasure = do
+    (mtrc, val) <- genBirthdateMeasure
+    Measure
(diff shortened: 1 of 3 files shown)


test/ModelTest.hs
```diff
@@ -443,15 +443,57 @@ test_14b = Test.stepsWithTransaction "test_14b" $ \step cn2 -> do
...
+    _ <- runReaderT
+        (do
+            [rcrd] <- lookupVolumeRecords vol
+            removeRecord rcrd)
+        aiCtxt
+    step "Then one doesn't see the record on the volume"
+    -- TODO: duplicates volumJSONField "records"
+    rs <- runWithNoIdent cn2 (lookupVolumeRecords vol)
+    fmap recordRow rs @?= []
+
```
kanishka-azimi: A little lengthy. I might break it up into multiple tests in the future. It's generally trying to capture "managing" a record.

test/ModelTest.hs
```diff
@@ -644,6 +691,13 @@ mkContainer v mRel mDate = do
            , containerRow = (containerRow c) { containerDate = mDate }
            })
 
+defineVolumeParticipantMetrics :: (MonadDB c m) => Volume -> [Metric] -> m ()
+defineVolumeParticipantMetrics vol metrics = do
+    -- should use add category + remove metric for each not used
+    forM_
+        metrics
+        (\m -> removeVolumeMetric vol (metricId m))
```
kanishka-azimi: Probably define this in the library.

test/ModelTest.hs
```diff
@@ -408,6 +409,7 @@ test_13 = Test.stepsWithTransaction "test_13" $ \step cn2 -> do
     rid <- runReaderT
          (do
               v <- addVolumeSetPrivate aiAcct
+              _ <- addVolumeCategory v (categoryId participantCategory)
```
kanishka-azimi: There is no safeguard in the system requiring metrics be attached to a volume before a record is defined. Seems bad.

kanishka-azimi:
My desire to use one test here is a result of there being no fixtures to start with, yet.

test/Model/Factories.hs
```diff
@@ -391,17 +391,27 @@ genGenderMeasure :: Gen (Metric, BS.ByteString)
 genGenderMeasure =
     pure (participantMetricGender, "Male")
 
-genParticipantMetricValue :: Gen (Metric, BS.ByteString)
-genParticipantMetricValue =
-    Gen.choice [genBirthdateMeasure, genGenderMeasure]
+-- TODO: genCreateMeasures -- will ensure each measure metric is distinct
+-- genParticipantMetricValue :: Gen (Metric, BS.ByteString)
```
chreekat: commented code identical to un-commented code

test/ModelTest.hs
```diff
@@ -433,7 +436,7 @@ test_14 = Test.stepsWithTransaction "test_14" $ \step cn2 -> do
     step "Then the public can't see the restricted measures like birthdate"
     (participantMetricBirthdate `notElem` (fmap measureMetric . getRecordMeasures) rcrdForAnon) @? "Expected birthdate to be removed"
 
-test_14b :: TestTree
+test_14b :: TestTree -- TODO: should record tests only test viewing under recordslot?
```
chreekat: Should records only be viewable as part of a slot? We should enforce that statically if so, allowing us to moot your TODO.

test/ModelTest.hs
chreekat: This seems very magical. What is begin "defined" by a SQL `DELETE` statement?

test/ModelTest.hs
```diff
@@ -443,15 +446,59 @@ test_14b = Test.stepsWithTransaction "test_14b" $ \step cn2 -> do
...
 
+test_14c :: TestTree
+test_14c = Test.stepsWithTransaction "test_14c" $ \step cn2 -> do
+    step "Given a volume record"
+    (aiAcct, aiCtxt) <- addAuthorizedInvestigatorWithInstitution' cn2
+    -- TODO: should be lookup auth on rootParty
+    (_, vol) <- runReaderT
+         (do
+              vol <- addVolumeWithAccess aiAcct
+              someMeasure <- Gen.sample genCreateGenderMeasure
```
chreekat: This should explicitly be a literal string, since you are changing it and checking that the change actually happened. Otherwise, how do you know that `sample` didn't return "Female" to begin with, creating a false negative test result? (I know that the generators aren't actually *generating* right now, but still.)

Part of the precondition should be checking that the measure datum is what you explicitly set it to, also.

```
setLiteral "foo"
checkRecord ?= "foo"
modifyRecord "baz"
checkRecord ?= "baz"

test/ModelTest.hs
chreekat: :+1:

test/ModelTest.hs
kanishka-azimi: No, can view them as part of a volume without any slot association.

test/ModelTest.hs
kanishka-azimi: probably a mistake here. can't remember if I meant `forM addVolumeMetric` or `addCategory + forM remove`

test/Model/Factories.hs
kanishka-azimi: I don't understand this comment.

chreekat: Mainly, lines 395–397 look like they can be deleted. They also happen to be identical to lines 394–396.

chreekat: Wait, I'm sorry. What I said doesn't make any sense.

kanishka-azimi: You were close enough. most of the commented out lines needed to be deleted. I only wanted to keep "TODO: genCreateMeasures"

test/ModelTest.hs
kanishka-azimi: I'm amending test_14b to handle the precondition assurance. One should be able to assume setting the value works by the time we reach 14c.

kanishka-azimi: BTW, since the metrics have their suggested values as part of their definitions, we can start use the suggested values in the generators later if desired.

kanishka-azimi:
When you get a chance, please indicate whether this version is good enough, after considering 14b and 14c.

chreekat:
lgtm

PR #608, opened 2018-06-30
Refactor partyJSON

kanishka-azimi:
I am about to add to this, but I want to merge this change first.

This moves towards how I want our json generation to work in the future.

PR #634, opened 2018-07-09
WIP: ToJson FormattedParty

chreekat:
## WIP. Todo:
- [ ] golden test against /api/party
- [ ] doctest for toJson

## Description

While reviewing #608, I felt like there had to be some way to do FormattedParty's ToJson using existing tools. I eventually found what I was looking for.

## Motivation and Context

Idiomatic code, yay

## How Has This Been Tested?

Ghci (see doctest)

## Types of changes

- [x] Refactor

## Checklist:

- [ ] I am sick of writing content for this PR, let's get this over with.

### DIFF
--- a/src/Model/Party.hs
+++ b/src/Model/Party.hs
@@ -94,17 +94,20 @@ data FormattedParty = FormattedParty
     , fpyAuthorization :: !(Maybe Permission)
     } deriving (Generic)
 
+-- | ToJSON options for FormattedParty
+formattedPartyToJsonOptions :: JSON.Options
+formattedPartyToJsonOptions = JSON.defaultOptions
+    { JSON.omitNothingFields = True
+    , JSON.fieldLabelModifier = (\(x:xs) -> toLower x : xs) . drop 3
+    }
+
 -- |
 -- >>> encode (FormattedParty 3 "foo" (Just "bar") Nothing Nothing Nothing Nothing Nothing Nothing Nothing)
 -- "{\"sortname\":\"foo\",\"id\":3,\"prename\":\"bar\"}"
 --
 instance JSON.ToJSON FormattedParty where
-    toEncoding =
-        JSON.genericToEncoding
-            JSON.defaultOptions
-                { JSON.omitNothingFields = True
-                , JSON.fieldLabelModifier = (\(x:xs) -> toLower x : xs) . drop 3
-                }
+    toJSON = JSON.genericToJSON formattedPartyToJsonOptions
+    toEncoding = JSON.genericToEncoding formattedPartyToJsonOptions
 
 partyRowJSON :: JSON.ToObject o => PartyRow -> JSON.Record (Id Party) o
 partyRowJSON PartyRow{..} = JSON.Record partyId $
--- a/src/Model/Party.hs
+++ b/src/Model/Party.hs
@@ -30,6 +30,7 @@ import Control.Applicative ((<|>))
 import Control.Exception.Lifted (handleJust)
 import Control.Monad (guard)
 import qualified Data.ByteString as BS
+import Data.Char (toLower)
 import Data.Int (Int64)
 import Data.List (intercalate)
 import Data.Maybe (isNothing, fromMaybe)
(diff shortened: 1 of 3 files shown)


chreekat:
Just going to finish running tests, then push the (not run) doctest..

src/Model/Party.hs
```diff
@@ -89,21 +91,12 @@ data FormattedParty = FormattedParty
...
-            <> "orcid" `JSON.omitIfNothing` fpyOrcid
-            <> "affiliation" `JSON.omitIfNothing` fpyAffiliation
-            <> "url" `JSON.omitIfNothing` fpyUrl
-            <> "institution" `JSON.omitIfNothing` fpyInstitution
-            <> "email" `JSON.omitIfNothing` fpyEmail
-            <> "permission" `JSON.omitIfNothing` fpyPermission
-            <> "authorization" `JSON.omitIfNothing` fpyAuthorization)
+    toEncoding =
+        JSON.genericToEncoding
+            JSON.defaultOptions { JSON.omitNothingFields = True }
```
kanishka-azimi: not right, needs to remove "fpy"

src/Model/Permission/Types.hs
```diff
@@ -39,7 +41,8 @@ data Permission
             Enum,
             GHC.Arr.Ix,
             Bounded,
-            Data.Typeable.Internal.Typeable)
+            Data.Typeable.Internal.Typeable,
+            Generic)
```
kanishka-azimi: `Data.Aeson.Types.toJSON . fromEnum` - wonder if it uses this or something else like show.

src/Model/Party.hs
chreekat: np. Fixed

src/Model/Party.hs
```diff
@@ -89,21 +92,19 @@ data FormattedParty = FormattedParty
...
-            <> "institution" `JSON.omitIfNothing` fpyInstitution
-            <> "email" `JSON.omitIfNothing` fpyEmail
-            <> "permission" `JSON.omitIfNothing` fpyPermission
-            <> "authorization" `JSON.omitIfNothing` fpyAuthorization)
+    toEncoding =
+        JSON.genericToEncoding
+            JSON.defaultOptions
+                { JSON.omitNothingFields = True
+                , JSON.fieldLabelModifier = (\(x:xs) -> toLower x : xs) . drop 3
+                }
```
kanishka-azimi: need toJSON - 

```
If on the other hand you wish to customize the generic decoding, you have to implement both methods:

customOptions = defaultOptions
                { fieldLabelModifier = map toUpper
                }

instance ToJSON Coord where
    toJSON     = genericToJSON customOptions
    toEncoding = genericToEncoding customOptions
```

chreekat:
I typo'd FormattedParty to FormattedParth, which I then "fixed" to FormattedPath.  Wow.

src/Model/Permission/Types.hs
chreekat: I presume Generic doesn't rely on Enum or Show... just GHC magic that provides String versions of the constructor names.

chreekat:
Good idea for that regression test... something is definitely wrong!

I am thinking it might be better to put this aside for now; it wasn't on the agenda. I'll keep going through with PR reviews first.

src/Model/Permission/Types.hs
kanishka-azimi: I wrote this comment because I was unclear why Generic was added here. There is already an explicit ToJSON instance below. Is `ToJSON FormattedParty` is using the explicit instance?

chreekat: While I could try taking it out, I presume it is required in in order to derive Generic for FormattedParty. But you have raised a good question: When FormattedParty's ToJson is derived via its Generic instance, is it using Permission's Generic instance or its ToJson instance?

kanishka-azimi:
Merge once you have the rebased, understood the behavior on Permission ToJSON, and tried out the smoke test.

PR #636, opened 2018-07-09
Switch from Maybe to Loaded for partySiteAccess field

kanishka-azimi:
## Description

In response to PR feedback, start implementing a more explicit type to denote fields that are not always loaded.

## Motivation and Context

Loaded type makes the implementation more self-documenting.

## How Has This Been Tested?

yes

## Types of changes

## Checklist:

- [x] I am sick of writing content for this PR, let's get this over with.

### DIFF
--- a/src/Model/Party.hs
+++ b/src/Model/Party.hs
@@ -1076,7 +1076,7 @@ findParties pf = do
                     vurl_a6R3W)
                  (do { cm_a6R44 <- vemail_a6R3X;
                        Just (Account cm_a6R44) })
-                 (maybe NotLoaded Loaded site))
+                 (Loaded site))
               Nothing
               ident)
       rows)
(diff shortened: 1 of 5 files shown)


chreekat:
Previously, I wasn't sure if you meant the *Party* was/wasn't loaded, or just the SiteAccess field within the Party. Now, I'm assuming it's the latter. But, what does it mean for a SiteAccess to be "loaded"?

kanishka-azimi:
> But, what does it mean for a SiteAccess to be "loaded"?

I don't know how to word this any differently. Some queries join the data in and load that field, some don't.

src/Model/Party.hs
```diff
@@ -1076,7 +1076,7 @@ findParties pf = do
                     vurl_a6R3W)
                  (do { cm_a6R44 <- vemail_a6R3X;
                        Just (Account cm_a6R44) })
-                 site)
+                 (maybe NotLoaded Loaded site))
```
chreekat: >  Some queries join the data in and load that field, some don't.

Ahh, I see. in that case, though, isn't the value `Loaded` in both cases here? The data is joined and looked for, after all. If it still doesn't exist, that suggests a `Loaded (Maybe a)`

chreekat:
ping, in case clicking 'approve changes' didn't already send a ping :)

PR #639, opened 2018-07-17
Minor refactor cleanup

chreekat:
## Description

Starting with implementing old versions in terms of new ones.

Should we actually leave things this way? The only difference between the implementations is the reliance on Config. When we remove that, we'll have to rewrite code either way.

## How Has This Been Tested?

Just unit tests, so far. I didn't look at the changes to hard, assuming the new versions are faithful reimplementations of the old versions. We should smoketest.

## Types of changes

- [x] Refactor

## Checklist:

- [ ] I am sick of writing content for this PR, let's get this over with.


src/Store/Service.hs
```diff
@@ -116,45 +115,19 @@ initStorage2 (Right StorageLocationConfig {..}) initTc = do
...
+    mkLocConf conf
+        | Just down <- conf C.! "DOWN" = Left down
+        | otherwise = Right StorageLocationConfig
+            { storageLocTemp = conf C.! "temp"
+            , storageLocMaster = conf C.! "master"
+            , storageLocUpload = conf C.! "upload"
+            , storageLocCache = conf C.! "cache"
+            , storageLocStage = conf C.! "stage"
+            , storageLocFallback = conf C.! "fallback"
+            }
```
kanishka-azimi: where function under initService

src/Store/Transcoder.hs
```diff
@@ -48,32 +48,11 @@ initTranscoder2 tconf@TranscoderConfig {..} = case (transcoderHost, transcoderDi
...
-                ++ err)
-  where
-  host = conf C.! "host"
-  dir = conf C.! "dir"
-  mount = conf C.! "mount"
+initTranscoder conf = initTranscoder2 TranscoderConfig
+    { transcoderHost = conf C.! "host"
+    , transcoderDir = conf C.! "dir"
+    , transcoderMount = conf C.! "mount"
+    }
```
kanishka-azimi: same as above

kanishka-azimi:
> We should smoketest.

Starting once in local development should be sufficient. A very cautious approach would also start it once on staging and perform an upload.

src/Store/Service.hs
chreekat: Ah, nice, will do

kanishka-azimi: Although, I rarely see functions in a where clause as justified. Usually a private function in a small module accomplishes the same with less indentation and syntax noise.

chreekat: Yeah, I agree.

PR #641, opened 2018-07-19
Refactor with HLint

chreekat:
1. Browse http://databrary.github.io/databrary/hlint.html
2. `$ hint='Redundant do' # for example`
3. `$ time vim -q <(hlint -j -o "$hint" src) +'sil cfdo sil !hlint -o "'"$hint"'" --refactor --refactor-options=-i %' +q`
4. Check, run tests
5. `$ git commit -am "Refactor: $hint"`
6. Repeat until satisfaction is achieved.

* [x] Unit tests
* [ ] nix-build (running)
* [x] smoketest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ repl: ; $(nix-shell) --run ./ghci-databrary.sh
 repl-test: ; $(nix-shell) --run 'cabal repl test:discovered'
 .PHONY: repl-test
 
+## Run configure for repltastic tasks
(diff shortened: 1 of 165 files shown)


chreekat:
I didn't use hlint's --serialise option because I got some funny error when I tried to use it. But I should try again, with restricted sets of hints applied.

chreekat:
Nix build failed because of two timeouts, which are totally unrelated. I need a new computer.

PR #642, opened 2018-07-24
Introduce requestVolume and Permissioned type

chreekat:
Rather than hacking around getVolume or using lookupVolume directly, tests can use requestVolume to get at the model data with particular permissions.

This makes concrete a common pattern in accessing objects. It will be copied for Slots, Containers, and maybe other things?

I added some incidental changes:
* move lookupContainerSlot to the library
* some tweaks to dev tools (more, unmerged changes are queued up)
* ruthless dead code removal

### DIFF
--- a/src/Model/Permission.hs
+++ b/src/Model/Permission.hs
@@ -12,6 +12,9 @@ module Model.Permission
   -- * Checking permissioned objects
   , checkPermission
   , PermissionResponse (..)
+  -- * New
+  , requestAccess
+  , Permissioned (Permissioned)
   ) where
 
 import Data.Monoid ((<>))
@@ -19,6 +22,28 @@ import Data.Monoid ((<>))
 import qualified JSON
 import Model.Release.Types
 import Model.Permission.Types
+
+-- | Represents a permissioned object.
+data Permissioned a = Permissioned
+    { unsafeAccess :: a
+    , grantedPermission :: Permission
+    }
+
+-- | How to get access to a permissioned object. It's not a great design, but it
+-- makes a concrete concept out of an existing pattern in the codebase. A better
+-- design could perhaps couple the access request to the action that needs the
+-- access.
+requestAccess
+    :: Permission
+    -- ^ Requested permission
+    -> Permissioned a
+    -- ^ object
+    -> Maybe a
+    -- ^ Maybe the unwrapped object
+requestAccess requestedPerm obj =
+    if requestedPerm >= grantedPermission obj
+    then Just (unsafeAccess obj)
+    else Nothing
 
 -- |Level at which things become visible. ; TODO: use this somewhere?
 -- permissionVIEW :: Permission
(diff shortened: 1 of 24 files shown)


src/Model/Permission.hs
```diff
@@ -20,6 +23,28 @@ import qualified JSON
 import Model.Release.Types
 import Model.Permission.Types
 
+-- | Represents a permissioned object.
+data Permissioned a = Permissioned
+    { unsafeAccess :: a
+    , grantedPermission :: Permission
```
kanishka-azimi: probably grantedRolePolicy

chreekat: Want to suggest some haddocks for the field? I'm still not sure what a 'role policy' is

kanishka-azimi: We discussed a little. You have to start by reviewing what a policy means in ABAC/XACML and what a role means in RBAC. I could commit to removing the word role everywhere and assuming ABAC concepts for now... I have been hesitating since I am hoping we limit the power of authorization to RBAC only soon.

kanishka-azimi:
Generally, if you observe the difference between the types named *Row vs * (e.g. VolumeRow vs Volume), you will observe that types like Volume already act as envelopes for "Permissioned" as well as joining other related data. Permissioned might be a typeclass that Volume implements... Regardless, that would be a bigger leap, so even if you agree with that note, this step helps reach there.

src/Model/Slot.hs
```diff
@@ -18,12 +20,34 @@ import Model.Audit
...
 
--- useTDB
-
+-- | Look up a Slot by its Id, gated by the running Identity's permission to view
+-- the Slot's Container's Volume. :)
 lookupSlot :: (MonadDB c m, MonadHasIdentity c m) => Id Slot -> m (Maybe Slot)
-lookupSlot (Id (SlotId c s)) =
-  fmap (`Slot` s) <$> lookupContainer c
+lookupSlot (Id (SlotId cont seg)) =
+  fmap (`Slot` seg) <$> lookupContainer cont
```
kanishka-azimi: if this were new code, I would want to avoid an fmap of fmap. its fine here

src/Model/Volume.hs
```diff
@@ -95,6 +61,41 @@ coreVolume = Volume
...
+    mkRequest =
+        maybe RequestDenied RequestResult . requestAccess requestedPerm
+
+-- | Lookup a 'Volume' by its Id, and wrap it in 'Permissioned'. The plan is for
+-- this to replace 'lookupVolume' entirely.
+lookupVolumeP
+    :: (MonadDB c m, MonadHasIdentity c m)
+    => Id Volume
+    -> m (Maybe (Permissioned Volume))
+lookupVolumeP = fmap (fmap wrapPermission) . lookupVolume
```
kanishka-azimi: fmap of fmap feels slightly too terse, but depends on the who is reading the code.

src/Model/Volume.hs
```diff
@@ -95,6 +61,41 @@ coreVolume = Volume
...
+  where
+    wrapPermission =
```
kanishka-azimi: I usually add a type signature.

PR #644, opened 2018-07-25
Permissioned Slot and Model.Access

chreekat:
Applies #642's treatment of getVolume to getSlot. Then it takes it one step further, combining the implementations and giving the concept a new name: Model.Access.

### DIFF
--- a/src/Controller/Slot.hs
+++ b/src/Controller/Slot.hs
@@ -25,6 +25,7 @@ import Controller.Volume (volumeIsPublicRestricted)
 import Controller.Web
 import HTTP.Path.Parser
 import Has (view, peeks)
+import Model.Access
 import Model.Asset
 import Model.AssetSegment
 import Model.AssetSlot
--- a/src/Controller/Slot.hs
+++ b/src/Controller/Slot.hs
@@ -55,11 +55,11 @@ getSlot
     -> Handler Slot
     -- ^ The slot, as requested (or a short-circuited error response)
 getSlot requestedPerm sId = do
-    res <- requestSlot requestedPerm sId
+    res <- accessSlot requestedPerm sId
     case res of
         LookupFailed -> result =<< peeks notFoundResponse
-        RequestDenied -> result =<< peeks forbiddenResponse
-        RequestResult s -> pure s
+        AccessDenied -> result =<< peeks forbiddenResponse
+        AccessResult s -> pure s
 
 -- | Look up a Slot and confirm that it is associated with the given Volume.
 --
--- a/src/Controller/Slot.hs
+++ b/src/Controller/Slot.hs
@@ -59,7 +59,7 @@ getSlot requestedPerm sId = do
     case res of
         LookupFailed -> result =<< peeks notFoundResponse
         AccessDenied -> result =<< peeks forbiddenResponse
-        AccessResult s -> pure s
+        Access s -> pure s
 
 -- | Look up a Slot and confirm that it is associated with the given Volume.
 --
--- a/src/Controller/Slot.hs
+++ b/src/Controller/Slot.hs
@@ -1,13 +1,13 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Controller.Slot
     ( getSlot
-    , getSlot'
+    , getVolumeSlot
     , viewSlot
     , slotDownloadName
     , thumbSlot
     ) where
 
-import Control.Monad (when, mfilter)
+import Control.Monad (when)
 import Data.Maybe (isJust)
 import Data.Monoid ((<>))
 import Network.HTTP.Types.Status (movedPermanently301)
@@ -21,7 +21,6 @@ import Controller.Angular
 import {-# SOURCE #-} Controller.AssetSegment
 import Controller.Container
 import Controller.Paths
-import Controller.Permission
 import Controller.Volume (volumeIsPublicRestricted)
 import Controller.Web
 import HTTP.Path.Parser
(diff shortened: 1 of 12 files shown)


src/Controller/Slot.hs
```diff
@@ -41,22 +41,43 @@ import Model.Volume
...
+    -- ^ Requested permission
+    -> Id Slot
+    -- ^ Slot to look up
+    -> Handler Slot
+    -- ^ The slot, as requested (or a short-circuited error response)
+getVolumeSlot volId requestedPerm sId = do
+    s <- getSlot requestedPerm sId
+    if volumeId (volumeRow (containerVolume (slotContainer s))) == volId
+        then pure s
+        else result =<< peeks notFoundResponse
```
kanishka-azimi: eventually, a type that encapsulates this like (Id Slot, Id Volume)

test/ModelTest.hs
```diff
@@ -282,30 +282,39 @@ test_11 = Test.stepsWithTransaction "test_11" $ \step cn2 -> do
...
+        )
+        aiCtxt
     step "Then the public can't view the container"
-    -- Implementation of getSlot PUBLIC
-    mSlotForAnon <- runWithNoIdent cn2 (lookupContainerSlot cid)
-    isNothing mSlotForAnon @? "expected slot lookup to find nothing"
+    mSlotForAnon <- runWithNoIdent
+        cn2
+        (requestSlot PermissionPUBLIC (containerSlotId cid))
+    mSlotForAnon @?= LookupFailed
```
kanishka-azimi: runWithNoIdent - move towards the pattern you identifid
(flip runReaderT) makeNoIdentCtxt (do
   ...
)

test/ModelTest.hs
```diff
@@ -282,30 +282,39 @@ test_11 = Test.stepsWithTransaction "test_11" $ \step cn2 -> do
...
-    -- Implementation of getSlot PUBLIC
-    Just slotForAnon <- runWithNoIdent cn2 (lookupContainerSlot cid)
+    slotForAnon <- runWithNoIdent
+        cn2
+        (requestSlot PermissionPUBLIC (containerSlotId cid))
     step "Then the public can't see protected parts like the detailed test date"
-    (encode . getContainerDate . slotContainer) slotForAnon @?= "2017"
+    case slotForAnon of
+        LookupFailed -> assertFailure "Lookup failed"
+        RequestDenied -> assertFailure "Access denied"
```
kanishka-azimi: usually expected ... , actual .... in message

src/Controller/Slot.hs
kanishka-azimi: ```data SlotLookupKey = DirectSlotId ... | VolumeSlotId ... ```

src/Model/Slot.hs
```diff
@@ -17,7 +18,38 @@ import Model.Identity
...
+requestSlot
+    :: (MonadDB c m, MonadHasIdentity c m)
+    => Permission
+    -> Id Slot
+    -> m (RequestResult Slot)
+requestSlot requestedPerm = fmap (maybe LookupFailed mkRequest) . lookupSlotP
+  where
+    mkRequest :: Permissioned Slot -> RequestResult Slot
+    mkRequest =
+        maybe RequestDenied RequestResult . requestAccess requestedPerm
```
kanishka-azimi: nested maybe feels a little spread out. I'd probably have one case or one nested case

src/Model/Slot.hs
```diff
@@ -17,7 +18,38 @@ import Model.Identity
...
+-- | Lookup a Slot by its Id, requesting the given permission.
+--
+-- NOTE: Intentionally implemented exactly like requestVolume. Implementations
+-- should be collected in a single module and merged.
+requestSlot
+    :: (MonadDB c m, MonadHasIdentity c m)
+    => Permission
+    -> Id Slot
+    -> m (RequestResult Slot)
+requestSlot requestedPerm = fmap (maybe LookupFailed mkRequest) . lookupSlotP
```
kanishka-azimi: I'd name the id slot parameter

src/Model/Slot.hs
```diff
@@ -17,7 +18,38 @@ import Model.Identity
...
+  where
+    mkRequest :: Permissioned Slot -> RequestResult Slot
+    mkRequest =
+        maybe RequestDenied RequestResult . requestAccess requestedPerm
+    --
+    lookupSlotP
+        :: (MonadDB c m, MonadHasIdentity c m)
+        => Id Slot
+        -> m (Maybe (Permissioned Slot))
+    lookupSlotP = fmap (fmap wrapPermission) . lookupSlot
```
kanishka-azimi: same feeling as before (:

test/ModelTest.hs
chreekat: Yeah that would be ideal. Can do.

src/Model/Slot.hs
chreekat: There's only one of them now. :)

chreekat:
> Overall, this PR and the last few PRs around controller.permission, getVolume, and getSlot, I would have aimed for less abstractions. [...] There is definitely a need for nicer abstractions. I just don't think you have enough context yet to know what those are.

You are correct that I don't have the whole picture/full context. If you think I'm going too far off the rails, I'm happy to revert changes.

When I am adding new features or fixing bugs, I try my absolute best to minimize diffs and introduce no new abstractions. In this case, though, my whole purpose is to refactor code under test with no change of behavior. I'm taking more liberties.

I'm trying to avoid perfectionism, too. In general, I don't think it's possible to keep the full picture of a software project in my head, and I don't want to rely on that hypothetical capability. The best I can do is create good abstractions for the specific piece I'm looking at. If I find an impedance mismatch upon moving to the next task, I can refactor again.  It is fun and easy to refactor with Haskell, and I want to take advantage of that.

Without a doubt, though, you have a better picture of things than I do! You've been dropping hints about the next pieces to look at (* versus *Row, *JSONField functions)... write more toehold tests, and I'll start ripping up those, too. ;)

Of course, one very real danger is creating a confusion of abstractions. Now that I've done getVolume and getSlot in a particular way, a newcomer might notice getContainer is implemented differently and not know why. I'd still rather move forward with piecewise refactorings, however. Comments and issues can keep track of works in progress, and even provide entry points for volunteers.

kanishka-azimi:
Don't revert anything. The new code is better, and feel free to keep applying this approach to similar functions.

I wrote my note for consideration when you start a new pass of refactoring (e.g. functions that create a volume, create a slot, etc).

> When I am adding new features or fixing bugs, I try my absolute best to minimize diffs and introduce no new abstractions. In this case, though, my whole purpose is to refactor code under test with no change of behavior.

It's funny. I work in the exact opposite style (:. When I am on a feature or bug, I feel like it focuses me on my future trajectory I want and I usually deal with the change at hand first, but quickly follow-up with huge refactorings. I guess we each have our own styles.

PR #651, opened 2018-08-03
Finish fixing issues with ghci-databrary.sh replacing ghci-databrary inline script

kanishka-azimi:
Original change was to make the nix scripts simpler and shorter.
This follow-up change is to fix issues we encountered with running node commands.

## Types of changes

- [ ] Style, layout change
- [x] Refactor
- [ ] Test
- [ ] Bug fix
- [ ] New feature
- [ ] Devops, scripts

### DIFF
--- a/ghci-databrary.sh
+++ b/ghci-databrary.sh
@@ -23,11 +23,12 @@ if [ ! -d "cracklib" ]; then
   mkdir -p cracklib
   cp -r /tmp/usr/share/cracklib/pw_dict* cracklib
 fi
-if [ ! -d "node_modules" ]; then
-  echo linking node_modules
-  # needed two dirnames instead of using ../.. because node's fs library choked otherwise  
-  ln -sfn $(dirname $(dirname $(which coffee)))/bin/node_modules node_modules # FIXME: Bryan
-fi
+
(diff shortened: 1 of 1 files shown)


kanishka-azimi:
cc: @chreekat  this might deal with some minor manual issues when upgrading node2nix

ghci-databrary.sh
```diff
@@ -23,10 +23,12 @@ if [ ! -d "cracklib" ]; then
...
 fi
-if [ ! -d "node_modules" ]; then
-  echo linking node_modules
-  ln -sf $(dirname $(which coffee))/../../../lib/node_modules node_modules # FIXME: Bryan
-fi
+
+# note: do this every time, so that node modules automatically updates when nix store version changes
+echo linking node_modules
+# note: needed two dirnames instead of using ../.. because node's fs library choked otherwise  
+ln -sfn $(dirname $(dirname $(which coffee)))/lib/node_modules node_modules # FIXME: Bryan
```
chreekat: I don't think `-n` is what you want. Best case, it does nothing; worst case, it creates `node_modules/node_modules` as the link.

Anyway, no change needed, this all needs to be defenestrated post-haste.

PR #653, opened 2018-08-03
Divorce runtime directory of ghci from runtime directory used by unit tests

kanishka-azimi:
Allow unit tests to run in an isolated directory with their own database instance and separate store directories. This isn't perfect, but its a step towards creating a cleaner separation. I used this same pattern when attempting to write an integration test suite that used haskell scripts and http client to drive the backend. You need to symlink files like databrary.cabal so that they are in the right location when the cabal command runs.

Feel free to object to this and request that we put this on hold for a while until we come up with the ideal approach.

## Motivation and Context

I really don't like having one database shared between running the application locally and running the (impure) unit tests. Tests shouldn't risk clashing with any manually created data.

My long term vision is to have three test suites:
- unit-pure defined in databrary.cabal
- unit-impure defined in pkgs/databrary-unit-impure/databrary-unit-impure.cabal
- inttest defined in pkgs/databrary-inttest/databrary-inttest.cabal

unit-impure will depend on the databrary library
inttest will depend on databrary library

unit-impure will eventually use a very small subset of ghci-databrary.sh
inttest will use most of ghci-databrary.sh

## How Has This Been Tested?

Ran unit tests from the new directory.

```
# first terminal
$ nix-shell --command "./init-db.psql.sh"
# second terminal
$ runtime-dirs/databrary-unit-impure/unit-impure-init-db.sh 
# third terminal
$ ./runtime-dirs/databrary-unit-impure/unit-impure-run-tests.sh 
```

## Types of changes

- [ ] Style, layout change
- [ ] Refactor
- [ ] Test
- [ ] Bug fix
- [ ] New feature
- [x] Devops, scripts


kanishka-azimi:
Will need solr directory whenever we attempt to turn on solr test again.

runtime-dirs/databrary-unit-impure/unit-impure-run-tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+# jsCssVersion.txt and empty web directory are still needed to start the backend
+# TODO: make flag to start backend only
+echo ":quit" | nix-shell --attr databrary-dev.env default.nix --command 'cd runtime-dirs/databrary-unit-impure && ../../ghci-databrary.sh'
+nix-shell --attr databrary-dev.env default.nix --command 'cd runtime-dirs/databrary-unit-impure && cabal -j test --test-options="--color always --hide-successes --timeout 2s"'
```
kanishka-azimi: One could optimize this.

A. Ghci-databrary.sh could skip running replace
B. The tests could run by using runghc instead of cabal build. (If we create very clean dependencies between modules, then cabal build's reuse might be faster than runghc).

kanishka-azimi:
Regarding such directories, I'm not sure if my ideal is to also create a directory called runtime-dirs/databrary and have developers change to this directory when they want to build and run the application, while leaving the current top level directory for building the library only, or whether I like the status quo.

chreekat:
+1 to independent runtime dirs. But couldn't all of this be implemented with config files/environment variables/command arguments? "databrary-nix-db" is already (mostly) a parameter, "dist" is definitely a parameter (--builddir), and all the databrary-specific things are controlled by the databrary config file.

chreekat:
I view the need for this as yet another indication we need to rewrite our build automation. If the lack of separation between runtime dirs is actively causing you trouble, though, go ahead and merge as a matter of expedience.

kanishka-azimi:
> But couldn't all of this be implemented with config files/environment variables/command arguments?

I am up for anything. We've had the status quo for many months now, so I just want to get the next proposal underway. I'll work on fleshing out your suggestion. This PR is biased towards the ideas I had in mind when implementing integration tests, which might have strictly needed ghci, instead of being able to use cabal built output as you indicate.

runtime-dirs/databrary-unit-impure/unit-impure-run-tests.sh
kanishka-azimi: TODO:
1. change ghci-databrary to not start repl when argument is provided
2. change test command to build in top level directory, but run tests here directly using dist/...
3. don't symlink databrary.cabal, src, or test

kanishka-azimi:
@chreekat Closer to what you noted.

I'll leave this open for a week. If we both think this is an improvement, then we can merge.

runtime-dirs/databrary-unit-impure/unit-impure-run-tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+# jsCssVersion.txt and empty web directory are still needed to start the backend
+# TODO: make flag to start backend only
+nix-shell --attr databrary-dev.env default.nix --command 'cd runtime-dirs/databrary-unit-impure && ../../ghci-databrary.sh --skip-ghci'
+nix-shell --attr databrary-dev.env default.nix --command 'cabal configure --enable-tests --datadir=. --datasubdir=. && cabal build discovered && cd runtime-dirs/databrary-unit-impure && ../../dist/build/discovered/discovered --color always --hide-successes --timeout 2s"'
```
kanishka-azimi: should probably let ghci-databrary.sh perform the cabal configure instead.

kanishka-azimi:
Not sure how much we gain from discovered being a test suite instead of executable.

ghci-databrary.sh
```diff
@@ -44,5 +49,7 @@ if [ ! -e "config/email" ]; then
   cp ${git_root_dir}/install/config.email config/email
 fi
 # rm -rf dist   # add this back when changing ffmpeg versions, c artifacts don't regenerate properly
-cabal configure --datadir=. --datasubdir=.
-cabal repl lib:databrary
+if [ "$skip_ghci" = false ] ; then
+    cabal configure --enable-tests --datadir=. --datasubdir=.
```
kanishka-azimi: Can't remember if you can use env variable to override datadir or not

chreekat:
I've been waiting to share these, but they do seem relatively stable, and relevant to this patch: https://github.com/databrary/databrary/pull/657

chreekat:
> This PR is biased towards the ideas I had in mind when implementing integration tests, which might have strictly needed ghci, instead of being able to use cabal built output as you indicate.

If you are using 'cabal repl', then it interacts with the dist dir. (Using ghci directly doesn't, but I don't think that's what you meant.) Having separate dist dirs for separate tasks sort of seems like the right way to go, unless we start using the new-* commands that do a better job of hashing contents based on configuration changes.

kanishka-azimi:
> Having separate dist dirs

I'll try to modify the cabal build command in run-tests to use a separate dist dir.

PR #654, opened 2018-08-04
Start of fake data generator utility

kanishka-azimi:
Start defining an executable, which a front end developer or back end developer might use after creating an empty test database on a test server.

## Motivation and Context
As we have long restricted our front end developer's ability to deploy to staging with a fully populated data set, let's at least provide some way to quickly generate fake data in test instances.

This will eventually grow to a few modules, and load the definition of the fake data from yaml files.

## How Has This Been Tested?
Ran it locally, as follows.

Start up database.
Run fake data insertion below.
```
$ nix-shell
> cabal build && dist/build/databrary-load-fake/databrary-load-fake
```
Then, start up your local instance as you normally would.

## Types of changes

- Devops, scripts

cc: @myicicle

### DIFF
--- a/app-databrary-load-fake/Main.hs
+++ b/app-databrary-load-fake/Main.hs
@@ -1,0 +1,107 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Main (main) where
+
+import Control.Monad.Trans.Reader
+import Data.Time
+import qualified Network.Wai as Wai
+
+import Has
+import Model.Audit (MonadAudit)
+import Model.Id
+import Model.Identity.Types
+import Model.Party (lookupSiteAuthByEmail)
+import Model.Party.Types
+import Model.Token.Types
+import Model.Volume.Types (Volume(..), VolumeRow(..), blankVolume)
+import Model.Volume (addVolume)
+import Model.VolumeAccess (setDefaultVolumeAccessesForCreated)
+import Service.DB
+
+main :: IO ()
+main = do
+    -- TODO add a test mode where this is wrapped in a transaction
+    ctx <- baseCtxt
+    Just auth <- runReaderT (lookupSiteAuthByEmail False "test@databrary.org") ctx
+    let superAdminAcct = siteAccount auth
+        superAdminCtx = setCtxtIdentInfo ctx auth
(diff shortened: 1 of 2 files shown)


kanishka-azimi:
There are two other methods that one can use to load fake data.

1. Start full backend, issue http requests
2. Drive controller level action handlers instead of invoking model directly

I might move closer to (2) next, to avoid introducing a redundant set of types that represent data generation requests.

app-databrary-load-fake/Main.hs
```diff
@@ -0,0 +1,107 @@
...
+main :: IO ()
+main = do
+    -- TODO add a test mode where this is wrapped in a transaction
+    ctx <- baseCtxt
+    Just auth <- runReaderT (lookupSiteAuthByEmail False "test@databrary.org") ctx
+    let superAdminAcct = siteAccount auth
+        superAdminCtx = setCtxtIdentInfo ctx auth
+    _ <- (flip runReaderT) superAdminCtx (do
+        addVolumeWithAccess superAdminAcct)
+    print ("Finished adding fake data." :: String)
```
chreekat: putStrLn

kanishka-azimi:
> I believe that replacing all of this with two lines of SQL would be a lot easier to understand, maintain

Agree on the on the other points, disagree on "maintain", very easy to drift. The test data hardcoded right now in 0.sql has already drifted from what a valid volume is.

Will ruminate on switching to sql. The biggest reason I think SQL won't be useful is because the data creation gets really complicated when you start considering assets and transcoding.

chreekat:
Should we execute 0.sql in the test suite?

kanishka-azimi:
> Should we execute 0.sql in the test suite?
It's executed when the database is created.. But we probably don't have enough tests against volume 1 and the user's created at the end of 0.sql to indicate the data needs to be updated.

chreekat:
https://projects.databrary.org/confluence/display/DD/Backend+Priorities?focusedCommentId=25362945#comment-25362945

kanishka-azimi:
If we go SQL, we should document how we captured the SQL, so that it is somewhat fast to capture again.

chreekat:
By the way, I don't consider sql a long term solution---just a better expedient solution.

PR #656, opened 2018-08-06
Garbage collect before copy closure instead

kanishka-azimi:
garbage collect before copy closure during deploy, instead of after installing new derivation on target server

## Motivation and Context
garbage collecting later on was probably deleting the previous derivation. that might have been safe, if there was only an executable involved, but the derivation also contains static assets that are loaded on demand into the web server. those static assets were probably deleted, leaving images and css files missing. (one can only observe this on the server, as most users will have the css and images cached)

## How Has This Been Tested?

ran a deploy to staging

## Types of changes

- Devops, scripts

### DIFF
--- a/build-package-copy-prep-prod
+++ b/build-package-copy-prep-prod
@@ -7,6 +7,7 @@ rcvr="$usr@databrary.home.nyu.edu"
 echo "building Databrary binaries..."
 # if /tmp is too small for major nix builds, create appropriate dir, then use TMPDIR=... build-package-copy-prep ...
 RESULT="$(nix-build --attr databrary --show-trace)"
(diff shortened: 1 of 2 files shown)


kanishka-azimi:
Test once more on deployment rehearsal server before merging

kanishka-azimi:
@chreekat In case you have any objections, feel free to comment, otherwise I will merge soon.

chreekat:
Nothing in the store is replaced; the nix store is immutable. The new package has a different hash. What *is* replaced is the symlink managed by nix-env. (That's why nix-env is considered the "imperative" interface to Nix.)

PR #657, opened 2018-08-08
Add repl/ghcid scripts

chreekat:
These are scripts I have been using to quick-load both the lib and tests. I have a tmux workflow that uses *three* running interpreters, running all three of these scripts.

See script comments for further docs.

To do?
* Remove Makefile targets that these scripts deprecate.

### DIFF
--- a/s/ghcid-lib
+++ b/s/ghcid-lib
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+#
+# Loads the library in ghcid
+#
+# Just to keep things interesting, ghcid is run within nix-shell, which both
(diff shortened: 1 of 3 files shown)


chreekat:
The other thing I don't like about these is they don't make mention of the fact that you need to be running the postgres script (`make db`) in the background.

kanishka-azimi:
entr is fine as a permanent dependency

I would rather have two levels of scripts. One at this automated level and one set of scripts slightly lower level that just let you invoke the command once. The second set of scripts are to allow for tinkering or playing with arguments passed in.

chreekat:
Since ghcid itself automates reload, even though I also prefer two levels in general, this case seems a bit futile.

PR #658, opened 2018-08-08
Update package.json and rebuild JS dependencies

chreekat:
This uses the package.json file from minyong-DBD-732-angular-upgrade, but doesn't do any of the extra Nix changes.

I've updated the instructions at https://projects.databrary.org/confluence/display/DD/How+to+upgrade+JS+dependencies to explain how to repeat what I've done.

@myicicle I'll leave it to you to see if this branch looks good, and merge if so. Let me know if there are any problems or questions, though!


chreekat:
Oh, by "looks good" I mean, can you load the site and see the new versions of dependencies.

Also, I left one thing out! You need to run the "./clean" script in the repository after checking out this branch. There is something broken somewhere still, so that the new deps don't automatically get added to the site. After you run ./clean, start the site and you should hopefully see the new deps.

kanishka-azimi:
> There is something broken somewhere still,

I suspect this should be easy to fix, just need to add a few lines to ghci-databrary.sh once we find the stale link.

PR #689, opened 2020-03-09
Added dropdown with comment about storing values

swapnaaiu:
## Types of changes

- Style, layout change
- Refactor
- Test
- Bug fix
- New feature
- Devops, scripts


TheoWolf:
The addition of variable is more complex and we are not going to deal with that for now

PR #709, opened 2021-11-01
Update Nix 2.0 install script URL to a working one

ngsankha:
The nix setup script uses an URL that does not work any more. Update the URL in this PR to the Nix releases site.

## Motivation and Context

Fixes the Nix install script.

## How Has This Been Tested?

Ran locally and it works

## Types of changes

- Bug fix


chreekat:
I haven't tried to run the Databrary setup scripts in a long while, but I can at least confirm that the updated url is correct. :)

chreekat (approved):

PR #716, opened 2023-03-13
codespell: config, workflow + fixing some typos

yarikoptic:
## Motivation and Context

My contribution to the field when I look at some related project ;)

## How Has This Been Tested?

nohow -- will wait for CI to tell (if there is any)

## Types of changes

- typos fixes and never make them happen again via a CI workflow

### DIFF
--- a/web/volume/spreadsheet.coffee
+++ b/web/volume/spreadsheet.coffee
@@ -307,7 +307,7 @@ app.directive 'spreadsheet', [
               Object.defineProperty @, v,
                 value: x
                 writable: true
-                configureable: true
+                configurable: true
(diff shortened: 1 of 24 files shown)


web/volume/spreadsheet.coffee
```diff
@@ -307,7 +307,7 @@ app.directive 'spreadsheet', [
               Object.defineProperty @, v,
                 value: x
                 writable: true
-                configureable: true
+                configurable: true
```
yarikoptic, line 310: not sure if might be introducing a bug or actually introducing one if whatever consuming this file is also having this typo.

